PR #32, opened 2016-05-14
pick time to output to catalog

xiansch:
Should the pick time be just the detect time and not detecttime + template start time - detect time (which I think is just template start time)?
Since plotting the events in catalog will trim to around the pick time.  So choosing the template start time as pick time doesn't cut the stream around where the actual detection was?

### DIFF
--- a/eqcorrscan/core/match_filter.py
+++ b/eqcorrscan/core/match_filter.py
@@ -662,11 +662,11 @@ def match_filter(template_names, template_list, st, threshold,
                 ev.comments.append(Comment(text=thresh_str))
                 ev.comments.append(Comment(text=ccc_str))
                 ev.comments.append(Comment(text=used_chans))
+                min_template_tm = min([tr.stats.starttime for tr in template])
                 for tr in template:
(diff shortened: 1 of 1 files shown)


calum-chamberlain:
Ah, no it shouldn't - but it shouldn't be what we had either.  I think it should be:
```python
min_template_tm = min([tr.stats.starttime for tr in template])
pick_tm = detecttime + (tr.stats.starttime - min_template_tm)
```
The `min_template_tm` line should be inserted before the `for tr in template` loop though so it isn't done too many times.  This should correct the `detecttime` for each channel based on the template start-time for that channel. 

Could you make that change and re-commit to this pull request?

xiansch:
changed the pull request.

calum-chamberlain:
Can you move the `min_template_line` to [before the loop](https://github.com/calum-chamberlain/EQcorrscan/pull/32/files#diff-02c274db86148c1776e8d8fa7be5c36eR665).  At the moment it gets reset for every channel in template when it doesn't need to be.

PR #38, opened 2016-07-04
Pull subspace in - do not merge until point met

calum-chamberlain:
@cjhopp has been working on a subspace branch, but it needs some refinement. Before we merge this in, the points below should be met:
- [x] subspace detector functions;
- [x] tests written for subspace detector (synthetic data and real data);
- [x] test different multiplexing technique;
- [x] optimize inner loop - cython?;
- [x] tutorial;
- [x] documentation.
- [x] a full, efficient looping method for multiple detectors;
- [x] plotting of the detector object (will only really be useful if not multiplexed);
- [x] Add network coincidence method.
- [x] test to confirm detections for multiplexed and non-multiplexed are the same.

~~- [ ] test for utils.clustering.min_sub_dimension~~


calum-chamberlain:
Now operates in an object-oriented sense using the Detector class.  Docs and tutorials written, just need to finalise tests.

calum-chamberlain:
Still needs:
- [ ] a full, efficient looping method for multiple detectors;
- [ ] plotting of the detector object (will only really be useful if not multiplexed;
- [ ] test against synthetic data;
- [ ] test to confirm detections for multiplexed and non-multiplexed are the same.

calum-chamberlain:
Non-multiplexed detection requires lagging, which requires delays to be tracked... Not sure how to do that.

calum-chamberlain:
Everything is added, merge once green!

calum-chamberlain:
Issue in creation of subspace detector - different on different systems.  I think this relates to the alignment of traces, will write an explicit test for this now and check.

PR #43, opened 2016-07-30
Match object

calum-chamberlain:
We should make use of simple object oriented things for matched-filtering, with a _Template_ object similar to the subspace.Detector object.  We could also then save template objects using h5py to maintain useful metadata meaning that we wouldn't have to keep track of them externally.

To do:
- [x] Develop a _Template_ object, with methods like _construct_, _read/write_, _detect_, use current functions **Needs a way to house many Templates and run them in tandem**;
- [x] Develop a _Family_ object to house detections from one template, which has io support, and plotting;
- [ ] Tutorial for this;
- [x] Docs;
~~- [ ] Mapping old templates in miniseed to Template objects.~~ Template() takes care of this.


calum-chamberlain:
Okay, this is nearly ready - testing is going to take a bit to debug, but if anyone @all want to run some of their work using this new API, go for it, let me know about bugs!

calum-chamberlain:
Using this branch in response to the [Kaikoura M7.5 earthquake](http://info.geonet.org.nz/display/home/2016/11/14/M7.5+Kaikoura+Earthquake%3A+Latest+updates), hopefully find some bugs and fix them!

calum-chamberlain:
`Tribe.write()` should merge all streams together before writing - currently very slow implementation, data are trimmed when read back in anyway.

calum-chamberlain:
This has been superseeded by using tar for io - this branch will remain for pyASDF development.

PR #44, opened 2016-07-30
Include GUIs for template creation and detection

calum-chamberlain:
Work has begun on template creation GUIs, which should access catalogues and plot waveforms to aid in decision-making regarding templates.  Some link to obspyck might be useful.
- [ ] Template generation gui;
- [ ] Parameter setting;
- [ ] Earthquake plotting;
- [ ] Spectral plotting for decision making for filtering;
- [ ] Interactive clustering.


calum-chamberlain:
I don't think this is worth thinking about for now.

PR #50, opened 2016-08-17
More integration of Nordic things

calum-chamberlain:
We should be able to read **all** the information from Nordic s-files, even if it doesn't have somewhere to go in the current obspy.core.event.Event object (e.g. full spectral info).
The following things should be implemented:
- [x] Read and write spectral information;
- [ ] Read focal mechanism solutions;
- [ ] Anything else?


calum-chamberlain:
Migrated nordic io to obspy

PR #58, opened 2016-10-10
Change coverage to codecov from coveralls

calum-chamberlain:
coveralls isn't picking up the coverage reports, and codecov is pretty. Lets see if this works.


calum-chamberlain:
Necessary bug-fix, not waiting for review.

PR #64, opened 2016-10-17
Bug-fix match_filter

calum-chamberlain:
0.1.4 introduced an issue for match_filter, which was attempted to be fixed in 0.1.5, but failed.  This pr has further tests and a patch.


calum-chamberlain:
Two users who previously reported bugs have reported this error fixed, and tests confirm fixed.  Will merge.

PR #67, opened 2016-11-03
Update plotting.py

cjhopp:
Rework detection_multiplot().

### DIFF
--- a/eqcorrscan/utils/plotting.py
+++ b/eqcorrscan/utils/plotting.py
@@ -860,15 +860,19 @@ def detection_multiplot(stream, template, times, streamcolour='k',
                                        86400)
                           for j in range(len(template_tr.data))]
         # Normalize the template according to the data detected in
-        normalizer = max(image.data[int((template_times[0] -
-                                        image_times[0]).
-                                        total_seconds() /
-                                        image.stats.delta):
-                                    int((template_times[-1] -
-                                         image_times[0]).
-                                        total_seconds() /
-                                        image.stats.delta)] /
-                         max(image.data))
+        try:
+            normalizer = max(image.data[int((template_times[0] -
+                                            image_times[0]).
+                                            total_seconds() /
+                                            image.stats.delta):
+                                        int((template_times[-1] -
+                                             image_times[0]).
+                                            total_seconds() /
+                                            image.stats.delta)] /
+                             max(image.data))
+        except ValueError:
+            # Occurs when there is no data in the image at this time...
+            normalizer = max(image.data)
         normalizer /= max(template_tr.data)
         axis.plot(template_times,
                   template_tr.data * normalizer,
--- a/eqcorrscan/utils/plotting.py
+++ b/eqcorrscan/utils/plotting.py
@@ -822,7 +822,8 @@ def detection_multiplot(stream, template, times, streamcolour='k',
     # Find only matching traces and sort by starttime with accompanying times
     traces_times = sorted([(tr, times[i]) for i, tr in enumerate(template)
                            if (tr.stats.station,
-                               tr.stats.channel[-1]) in stream_stachans], key=lambda x: x[0].stats.starttime)
+                               tr.stats.channel[-1]) in stream_stachans],
+                          key=lambda x: x[0].stats.starttime)
     traces, times = zip(*traces_times)
     template.traces = traces
     ntraces = len(template)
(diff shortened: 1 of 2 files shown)


cjhopp:
Hold off on this for a second or two...

calum-chamberlain:
Ace, so the process will be, make sure the tests all pass (they should), then I will give it a quick review and we can then merge. If you want to make any changes you can continue to commit to this branch and the pull request will update.

calum-chamberlain:
Also, could you make a note in the CHANGES.md file outlining (briefly) what you did so we can keep a track of the bug-fixes :)

cjhopp:
Aight. Should be good to go. I was getting intermittent errors, but was due to bad input data on my end.

calum-chamberlain:
Looks like a pep8 fail.

cjhopp:
Issue with the lambda...? And what's the issue with the last formatting string..?

calum-chamberlain:
Give us some links boyo.

cjhopp:
Lambda issue here: https://github.com/eqcorrscan/EQcorrscan/blob/fbd2ebd3281e9fa9e0fb90b32aab941f559aa5b3/eqcorrscan/utils/plotting.py#L823-L825

cjhopp:
String formatting here: [https://github.com/eqcorrscan/EQcorrscan/blob/fbd2ebd3281e9fa9e0fb90b32aab941f559aa5b3/eqcorrscan/utils/plotting.py#L883-L887](url)

cjhopp:
Must be better way to share links...

calum-chamberlain:
Link to the failed test! That looks like it will have failed because the line is too long... Does the test that failed say 'line too long'....
And there is a better way: [pep8 fail](https://ci.appveyor.com/project/calum-chamberlain/eqcorrscan-jsycv/build/1.0.180)

cjhopp:
Isn't that link available via 'Details' next to the pertinent test?

calum-chamberlain:
It is indeed, but the one that I looked at was an obvious pep8 fail - I thought you had something different? There are 12 test sessions run and I am too lazy to look at all of them.

cjhopp:
Let's try that again...

calum-chamberlain:
Looks like everything should pass - I think the codecov/patch might fail, but that's because we don't report coverage on plotting.  I will give this a quick review tomorrow. :+1:

eqcorrscan/utils/plotting.py
```diff
@@ -852,43 +853,42 @@ def detection_multiplot(stream, template, times, streamcolour='k',
...
-                      template_tr.data * normalizer,
-                      templatecolour, linewidth=1.2)
+        time = times[i]
+        time_dt = time.datetime
+        template_times = [time_dt +
+                          dt.timedelta((j * template_tr.stats.delta) /
+                                       86400)
+                          for j in range(len(template_tr.data))]
+        # Normalize the template according to the data detected in
+        normalizer = max(image.data[int((template_times[0] -
```
calum-chamberlain: Pretty sure there is a prettier way to write this, but this is fine.

eqcorrscan/utils/plotting.py
```diff
@@ -852,43 +853,42 @@ def detection_multiplot(stream, template, times, streamcolour='k',
...
+                                        image_times[0]).
+                                        total_seconds() /
+                                        image.stats.delta):
+                                    int((template_times[-1] -
+                                         image_times[0]).
+                                        total_seconds() /
+                                        image.stats.delta)] /
+                         max(image.data))
+        normalizer /= max(template_tr.data)
```
calum-chamberlain: You removed the try/except statement, any reason for this? I think it looks (from my comment) like it was necessary when there was no data present?

eqcorrscan/utils/plotting.py
```diff
@@ -852,43 +853,42 @@ def detection_multiplot(stream, template, times, streamcolour='k',
...
+                  templatecolour, linewidth=1.2)
         ylab = '.'.join([template_tr.stats.station,
                          template_tr.stats.channel])
         axis.set_ylabel(ylab, rotation=0,
                         horizontalalignment='right')
-        # axis.yaxis.set_ticks([])
+        axis.yaxis.set_ticks([])
     if len(template) > 1:
-        axes[len(axes) - 1].set_xlabel('Time')
+        axes[-1].set_xlabel('Time (s)')
```
calum-chamberlain: Based on the mdates formatter below, s isn't the correct unit.

eqcorrscan/utils/plotting.py
```diff
@@ -852,43 +853,42 @@ def detection_multiplot(stream, template, times, streamcolour='k',
...
+        axes[-1].xaxis.set_major_locator(mdates.SecondLocator())
+        axes[-1].xaxis.set_major_formatter(mdates.DateFormatter('%H:%M:%S'))
+        xaxis.set_major_formatter(mdates.DateFormatter('%H:%M:%S'))
     else:
         axis.set_xlabel('Time')
+        axis.xaxis.set_major_locator(mdates.SecondLocator())
+        axis.xaxis.set_major_formatter(mdates.DateFormatter('%H:%M:%S'))
     plt.subplots_adjust(hspace=0, left=0.175, right=0.95, bottom=0.07)
-    plt.xticks(rotation=10)
+    # plt.xticks(rotation=10)
```
calum-chamberlain: I had the rotation there so plot-labels didn't overlap, is that not an issue now?

eqcorrscan/utils/plotting.py
cjhopp: I didn't mess with any of the normalizing. But I think it's fine.

eqcorrscan/utils/plotting.py
cjhopp: So for plots with less than ~10 seconds the way it's written currently works well. Maybe it's too much to assume that templates will always be of a certain length (<3 sec or so), and will therefore be best plotted on axes of less than 10 seconds? The only reason it even came to mind is that Martha complained about the diagonal labels on a cumulative detection plot once.

cjhopp:
......I'm an idiot. In looking at the example .png you already have, I realized that the entire purpose is to allow multiple detections to be plotted on one trace.

So the way it was written was mostly fine, with maybe a minor tweak to the documentation and maybe in what the argument 'times' consists of...

PR #68, opened 2016-11-03
Update plotting.py

cjhopp:
Allow detection_multiplot to handle streams with fewer traces than templates.


calum-chamberlain:
This needs an example, leaving this until post-quake has calmed a bit.

- [ ] Example to show syntax.

eqcorrscan/utils/plotting.py
```diff
@@ -799,9 +799,7 @@ def detection_multiplot(stream, template, times, streamcolour='k',
     :param stream: Stream of data to be plotted as the background.
     :type template: obspy.core.stream.Stream
     :param template: Template to be plotted on top of the base stream.
-    :type times: list
-    :param times: list of times of detections in the order of the channels in
-                template.
+    :type times: list of detection times
```
calum-chamberlain: This looks like it is missing a doc-string, type should just be the type.

calum-chamberlain: Also needs an example.

cjhopp:
See if that looks better?

PR #91, opened 2017-04-27
SVD patch and small changes for memory debug

cjhopp:
Honestly, I forgot to commit the changes I made in match_filter to cjhopp-mem_spike-patch......sloppy.

Thoughts?

P.S. The changes to svd() require new subspace h5 files in the testing directory. Did you make those from events locally?


calum-chamberlain:
Ideally I would try and separate them out - maybe close this PR and then pick your commits? I'm not great at that though...

Yes I made them locally, is the information there to re-make them? Which test is it - is it just doing a compare between one on disk and one that it makes? If so you should re-make them using the test, and confirm that they are what you expect.  Or, if it's just naming things, then read the ones from disk in and change the naming to what you have now and write back?

PR #92, opened 2017-04-28
Subspace Freq domain statistic

cjhopp:
Rebase seemed to work....maybe.

Your original suggestion of reading in the test detectors, renaming them, and then re-writing them should work with some minor transposing of matrices. I'll push those in a bit.


cjhopp:
Hey @calum-chamberlain, ever seen the current Travis fail for 3.5 on Linux? Some strange issue with an invalid resize at the end of channel_loop....Did I break it???

calum-chamberlain:
@cjhopp No I have not... Curious, going to try and re-run, and will try the tests locally also.
@d-chambers have you had a look at this? We were struggling to reason why @cjhopp had to decimate not from the zeroth index of the detection statistic in the frequency domain - it kind of makes sense, but I can't describe it - any insight or recommendations?

cjhopp:
To tag on to that a bit, @d-chambers, I'm curious if you've been able to implement the aliasing in the frequency domain described by Harris (and, if so, were the time savings worth the trouble?). We've had a crack at it, but it isn't producing the correct results.

Dear Panda, No hard feelings. -Chet

d-chambers:
hey @cjhopp, in detex I didn't implement the freq. domain decimation. I think when I got to that point I was in a hurry to graduate so I just had to get the code working. 

Happy to take a look at your implementation though. Will try to add a few comments when I get a chance.

d-chambers:
@cjhopp, where are you doing frequency domain decimation? I think I am missing it.

cjhopp:
@d-chambers  Sorry, I should have been more clear with that question! I didn't end up putting it in this PR as we couldn't make sense of it. From my point of view, the only extra step would be adding something like this: `num_cor.reshape(num_cor.shape[0], Nc, M).sum(axis=1)` [here](https://github.com/eqcorrscan/EQcorrscan/blob/b1935e33d990dbef6ffbd6b9dbc80160128e2c07/eqcorrscan/core/subspace.py#L607) where `M` should be `len(template) + len(data) - 1`.

Hopefully, I can get another crack at it next week, but if you have any thoughts about how best to do it, I'm all ears.

cjhopp:
[This](https://github.com/eqcorrscan/EQcorrscan/blob/571261c03baab256277955e8ddac1788cb4f8aea/eqcorrscan/core/subspace.py#L611-L633) was as far as the implementation got.

d-chambers:
Thanks for pointing that out, I am fairly new to a lot of the github functionality so the extra help is appreciated. 

I left you two general comments in the review. 

Interesting frequency domain normalization [here-ish](https://github.com/eqcorrscan/EQcorrscan/blob/b1935e33d990dbef6ffbd6b9dbc80160128e2c07/eqcorrscan/core/subspace.py#L607). Did you profile against time domain implementations (with bottleneck moving variance for example?) or maybe [this issue](https://github.com/kwgoodman/bottleneck/issues/164) makes it better to avoid bottleneck so float 32 can be used? 

Regarding the frequency domain decimation; I haven't been able to try it out yet. I have a deadline today and I am on vacation next week, so it may be awhile before I can really dig into it. it is probably worth going ahead without it, we can also come back to it (it shouldn't affect any APIs).  Might be a reshape issue? Maybe the order should be 'f' rather than "c"?

I sent @calum-chamberlain a notebook awhile ago with my current thoughts on the subspace implementation and general UX concepts I am partial to, I will send it to you as well. I like the idea of using [xarray](http://xarray.pydata.org/en/stable/) to store all data then using broadcasting to avoid python loops. I am currently trying to get a prototype working. I am not sure how hard it would be to implement such a data structure in eqcorrscan, I think calum is exploring that space.

cjhopp:
@d-chambers Extremely useful comments, thanks! do_ffts() should definitely be private and I wasn't aware that the ffts were no longer sensitive to power-of-2 lengths. 

The frequency domain normalization you reference was really just an attempt to recreate Harris' approach literally. I haven't profiled a time domain approach but I don't have any attachment to the way it is currently so I'll give bottleneck a shot tomorrow.

Re: the frequency decimation, I hadn't considered the array order, so it may very well be a C/Fortran order issue. I'll give that a try. I agree that it isn't imperative right away, though.

Also, thanks for sending over your notebook. I'll give it a closer read first thing in the morning. Calum is close to pushing some internal changes involving xarray, I believe.

Hope you have a nice bit of vacay this week!

calum-chamberlain:
No xarray, still a ways from that guys, sorry! My branch is painfully misnamed.

Thanks though guys, really awesome to have more eyes and knowledge on things!

Calum Chamberlain
Sent from mobile.

cjhopp:
Benchmarks for _do_ffts()
```
# Next power of 2
%timeit _do_ffts(detector, stream, Nc)
1 loop, best of 3: 1.06 s per loop 
# next_fast_len()
%timeit _do_ffts(detector, stream, Nc)
1 loop, best of 3: 388 ms per loop
# next_fast_len() and rffts
%timeit _do_ffts(detector, stream, Nc)
1 loop, best of 3: 202 ms per loop
```
Then, for the detection statistic calculation:
```
# Full ifft
%timeit _det_stat_freq(det_freq, data_freq_sq, data_freq, fft_vars[3], Nc, fft_vars[4], fft_vars[5])
1 loop, best of 3: 837 ms per loop
# Using rfft
%timeit _det_stat_freq(det_freq, data_freq_sq, data_freq, fft_vars[3], Nc, fft_vars[4], fft_vars[5])
1 loop, best of 3: 180 ms per loop
```
Pushing changes now. Thanks for the feedback, @d-chambers!

eqcorrscan/core/subspace.py
```diff
@@ -493,27 +492,25 @@ def _detect(detector, st, threshold, trig_int, moveout=0, min_trig=0,
...
     outtic = time.clock()
-    if debug > 0:
-        print('Computing detection statistics')
     # If multiplexed, how many samples do we increment by?
     if detector.multiplex:
-        inc = np.uint32(len(detector.stachans))
+        Nc = len(detector.stachans)
     else:
-        inc = np.uint32(1)
-    # Stats must be same size for multiplexed or non multiplexed!
```
calum-chamberlain: Maintain indent for comment here.

eqcorrscan/core/subspace.py
```diff
@@ -586,6 +582,41 @@ def _detect(detector, st, threshold, trig_int, moveout=0, min_trig=0,
     return detections
 
 
+def _do_ffts(detector, stream, Nc):
```
calum-chamberlain: Can we have a doc-string for this please? I like to have them even for internal functions so we know what's going on (I probably don't have them for everything though!)

cjhopp: On it.

eqcorrscan/core/subspace.py
cjhopp: Comment removed in some recent commit.

calum-chamberlain:
Thanks @cjhopp - worth just merging it seeing as no changes made would have affected the code (all doc-strings).  Press the button and update #100!

PR #93, opened 2017-04-30
Patch dayproc

calum-chamberlain:
This PR copes with a problem specific to GeoNet data, where trace starttimes are shifted by less than a sample, resulting in day-long data beginning on the previous day compared to standard data.

This should check whether traces start within 1 sample of the next day, then set the start time for process to that day, rather than potentially setting starttime to the wrong (previous) day.


calum-chamberlain:
You're right, that last commit unsets start-time in _group_detect.  I don't think it gets set elsewhere (e.g. for lag-calc). Unsetting it should force it to go through that loop (or rather setting it to `None`).

calum-chamberlain:
Ah no, it definitely does get called in group_process
Will try and fix.

calum-chamberlain:
@cjhopp whadya reckon? I should probably add a little test to check that it raises the error expected, but otherwise?

PR #94, opened 2017-05-22
Update internals to vectorised SciPy

calum-chamberlain:
This PR is a work in progress update to the internals of match-filter. This PR will:

- Remove the dependancy on OpenCV (for simpler install);
- Change to SciPy internals for more flexibility (will compute fewer ffts);
- Compute cross-correlations in a vectorised routine for better efficiency for multiple templates;

To be done:
- [x] Remove old functions, or allow them to be used with an optional argument to match_filter;
- [x] Work on memory constraints for large, parallel problems - Dask/temporary files?
- [x] Run more tests on large datasets.
~~- [ ] Implement multi-threaded fft? (pyfftw)~~
- [x] Only use float64 for move_std (use float32 for other calculations to conserve memory)
- [x] Remove dependancy on openCV for other functions - rewrite normxcorr2 as a SciPy correlation?
- [x] Check whether _spike_test is needed now, if not, remove because it's really slow!

**EDIT**:

This PR has evolved into a complete overhaul of the internals of EQcorrscan - Scipy was tested and found to be memory inefficient when run in parallel, also memory profiler issues with SLURM have forced a desire to move away from python multiprocessing.
All this has led to the development of C internals using FFTW and openMP.


calum-chamberlain:
Note to @cjhopp and @d-chambers this is not yet ready for a full review, but @cjhopp may want to test this on PAN to see if this gets around the memory issues.  So far I have run the tests for synthetic data, but not for real data (haven't had the time at JpGU), so wait for CI etc.

cjhopp:
Fingers crossed. I'll have a look and see in the next day or so.

calum-chamberlain:
ci fails look like install issues, but tests don't run at the moment on other machines - running into the bottleneck issue here: [bottleneck: 164](https://github.com/kwgoodman/bottleneck/issues/164)

eqcorrscan/core/match_filter.py
```diff
@@ -3360,6 +3386,170 @@ def normxcorr2(template, image):
...
+    stream_mean_array = bottleneck.move_mean(
+        stream, template_length)[template_length - 1:]
+    stream_std_array = bottleneck.move_std(
+        stream, template_length)[template_length - 1:]
+    # Normalize and flip the templates
+    norm = ((templates - templates.mean(axis=-1, keepdims=True)) / (
+        templates.std(axis=-1, keepdims=True) * template_length))
+    norm_sum = norm.sum(axis=-1, keepdims=True)
+    stream_fft = np.fft.rfft(stream, fftshape)
+    template_fft = np.fft.rfft(np.flip(norm, axis=-1), fftshape, axis=-1)
```
d-chambers: [Numpy's flip](https://docs.scipy.org/doc/numpy/reference/generated/numpy.flip.html) was added in version 1.12.0 so we need to make sure and bump the version in the setup.py

eqcorrscan/core/match_filter.py
```diff
@@ -3360,6 +3386,170 @@ def normxcorr2(template, image):
...
+    #     xcorrs = []
+    #     for seed_id in seed_ids:
+    #         tr_xcorrs, tr_chans = dask.delayed(multi_normxcorr)(
+    #             templates=template_array[seed_id],
+    #             stream=stream.select(id=seed_id.split('_')[0])[0].data)
+    #         xcorrs.append(tr_xcorrs)
+    #     cccsums = dask.delayed(np.sum)(xcorrs, axis=0)
+    #     if compute:
+    #         cccsums.compute()
+    if cores is None:
```
d-chambers: Would it make sense to abstract the multiprocessing further up? I know several functions use something similar so maybe we could make a generic pool interface on the module level, then we could have persistent processes/threads in the pool so we wouldn't need to spin them up every time.

setup.py
```diff
@@ -72,21 +72,20 @@
...
-        install_requires = ['numpy>=1.8.0', 'obspy>=0.10.2',
+        install_requires = ['numpy>=1.8.0', 'obspy>=1.0.0',
                             'matplotlib>=1.3.0', 'joblib>=0.8.4',
-                            'scipy>=0.14', 'LatLon', 'h5py', 'cython',
-                            'bottleneck']
+                            'scipy>=0.18', 'LatLon', 'cython',
+                            'bottleneck', 'xarray']
     else:
-        install_requires = ['numpy>=1.8.0', 'obspy>=0.10.2',
+        install_requires = ['numpy>=1.8.0', 'obspy>=1.0.0',
```
d-chambers: as mentioned above we need at least np 1.12.0 because we are using flip

eqcorrscan/core/match_filter.py
```diff
@@ -51,6 +52,31 @@
 from eqcorrscan.core.lag_calc import lag_calc
 
 
+def _spike_test(stream, percent=0.99, multiplier=1e6):
```
d-chambers: Is this for avoiding [bottleneck: 164](https://github.com/kwgoodman/bottleneck/issues/164) ? It might be worth adding a bit in the doc string about why this function might be used.

calum-chamberlain: It's not, this was for an issue from openCV correlations - I'm not sure if it is needed for the scipy internals - I need to check that and if it's not needed I should remove it - it's really slow and at least one user has complained!

setup.py
calum-chamberlain: Thanks!

eqcorrscan/core/match_filter.py
calum-chamberlain: Obspy 1.0.3 does not play nice on Windows, but the current master does, I'm going to keep the appveyor running the osbpy master and travis running the current release until the next obspy release when appveyor should revert to the current obspy release.  I think it comes down to their pinning of matplotlib...?

calum-chamberlain:
So this is still really painful in terms of memory usage - the variables created within the parallel loop suck memory.
I also want to get rid of Python multiprocessing, because this is not very compatible with SLURM systems (for the type of memory profiling used on the NeSI PAN cluster a massive and un-real spike is registered, which kills code).
This is all pushing me towards full c++ internals - I have a cpp, parallel loop for openCV, but it isn't much faster than the vectorised SciPy is - it's still just using the openCV match_Template function, which results in doing far more fft's than necessary.  I'm going to try and write an fftw derived multi-correlation loop...
Watch this space

calum-chamberlain:
@d-chambers I did some work trying to reduce the memory used in these SciPy internals (which also gets around the use of bottleneck and associated nan errors), which resulted in my writing a C routine for moving window normalisation - I thought you might be interested?  

The C routine is slower than the bottleneck version for small cases (e.g. for a correlation vector of 8.64x10^(6) points and 3 templates the bottleneck case is 25x faster (0.02s v. 0.5s)), but is faster for large cases, where the speed ends up being noticeable anyway (for 8.65x10^(6) long vector and 300 templates bottleneck is 15s and the C version is 12s).

Most importantly though, there is almost no additional memory usage from my C version, whereas the bottleneck version generally at-least doubles memory usage because it:
 - creates two large arrays of means and std
 - needs to use 64 bit floats

As ever it's openly licensed and now installs on various machines without issue.  I think it could be made faster with a better way of updating the moving std, but I couldn't work it out.

Eventually I will write the whole correlation in C so that I can get away from python multiprocessing while still having an efficient way of using large machines or GPUs, but this means that these SciPy internals can be used without too much memory cost (meaning large problems are tractable). Open to comments though.

d-chambers:
Looks interesting.

I just got back from a conference and I am heading into the field for a few weeks so I doubt I will have much time to play with it but I might have an hour or so tomorrow. Where do you expose the C function in eqcorrscan? 

Introducing code that needs to be compiled makes the installation story much more complicated, but you are right, it may be necessary. A few thoughts though:

1. Did you look at @cjhopp frequency domain normalization implementation? I think he is just doing a unit boxcar convolution to get the moving sum then dividing by the number of samples to get the mean but I am not sure how he is doing the moving var or std (I didn't have a chance to look closely at it yet). Maybe it is fast enough and memory efficient enough?

2. Might be interesting to look at a [Numba](https://numba.pydata.org/) implementation. I have heard it supports compiling python into CUDA kernels (more on that [here](http://numba.pydata.org/numba-doc/0.33.0/cuda/index.html)) so it may be a quick way to get going on the GPU. 

Do you think the internals will need to be written in C to have a good parallelization on a cluster anyway?

calum-chamberlain:
Good luck in the field, hope the weather plays nice.

Agreed on the compilation making install more complicated, especially when trying to be system agnostic, but I do think that, for the cluster that we use, we can't use python multiprocessing because the memory profiler that the SLURM controller (job scheduler etc) uses records a memory spike (that isn't real), and the cluster maintainers are not willing to change the memory profiling system.  So I think the internals need to not use multiprocessing, (which I think Dask uses too, but I might be wrong).

RE your suggestions:
 - I didn't look at Chet's normalisation for this, but looking at it now, I think it suffers from similar memory issues for large datasets by computing a whole vector of means etc.
 - Good shout on the numba stuff, will add it to my reading list!

Thanks!

calum-chamberlain:
FFTW derived C routine written for cross-correlation (and time-domain to compare results to) written and tested locally.  Need to test elsewhere, the CI will likely fail because I haven't sorted compilation with fftw libraries.

Runs a little faster than the SciPy implementation (about 15% faster), but is a step on the way to using openMP for parallel workflows.

Need to:
 - [ ] sort out installs on CI,
 - [ ] update CHANGES.md
 - [ ] update install instructions
 - [ ] try using fftw 2D for templates, currently still doing too many ffts
 - [ ] write openMP loop around normxcorr_fftw

Once that lot is done, I think this will be finished for me - not that it is either Xarray or SciPy now...

eqcorrscan/core/match_filter.py
calum-chamberlain: So I have sped this up a lot, and it looks like it is needed to stabilise ffts.

eqcorrscan/tests/match_filter_test.py
```diff
@@ -78,56 +78,11 @@ def test_set_normxcorr2(self):
...
         expected_ccc = np.load(os.path.join(testing_path, 'test_ccc.npy'))
-        # We know that conda installs of openCV give a different results
-        # to source built - allow this and allow it to pass.
-        self.assertTrue((np.gradient(expected_ccc).round(2) ==
-                         np.gradient(ccc).round(2)).all())
-        if not (ccc == expected_ccc).all():
-            warnings.warn('The expected result was not achieved, ' +
-                          'but it has the same shape')
-
-    def test_perfect_template_loop(self):
```
calum-chamberlain: This is tested in the normxcorr2 tests as _template_loop has been removed.

eqcorrscan/tests/match_filter_test.py
```diff
@@ -368,6 +323,19 @@ def test_detection_extraction(self):
         self.assertEqual(len(detections), 4)
         self.assertEqual(len(detection_streams), len(detections))
 
+    def test_normxcorr(self):
```
calum-chamberlain: These data have large variations in amplitude late in the continuous data, which leads to floating point error accumulation when using float (32-Bit) in the C routine - overcome by using double (64-Bit float) internally within the C-routine.

eqcorrscan/tests/match_filter_test.py
```diff
@@ -788,14 +763,14 @@ def test_day_long_methods(self):
         # Aftershock sequence, with 1Hz data, lots of good correlations = high
         # MAD!
         day_party = daylong_tribe.detect(
-            stream=st, threshold=4.5, threshold_type='MAD', trig_int=6.0,
```
calum-chamberlain: This threshold was too low to produce meaningful results.

eqcorrscan/tests/tutorials_test.py
```diff
@@ -17,12 +18,19 @@
 from eqcorrscan.core.match_filter import read_detections
 
 
+slow = pytest.mark.skipif(
```
calum-chamberlain: This is so that we don't always have to run the tutorial tests - they rely on long data downloads, which appveyor is not so great at handling.

calum-chamberlain:
A quick note: tests are passing (despite the apparent fail on appveyor, something is wrong with the interface there), I'm going to work on upping test coverage, but could do with a review.  A lot of the changes in here are extra little patches, the main things that need looking at are:
- [ ] eqcorrscan/lib/multi_corr.c
- [ ] eqcorrscan/utils/correlate.py

A further note, I have deliberately not written the c functions with python bindings, because I would like them to be employed by other programs if wanted (e.g. matlab or other C funcs).
@cjhopp fancy having a little scan?  I'm going to get on with testing on PAN soon, but I think any issues there (memory problems) will result in a further PR with some other extra set of implementations.

calum-chamberlain:
Merged. Will work on openMP loop implementation (for SLURM applications) elsewhere.

PR #97, opened 2017-05-31
Debug failing lag_calc test

calum-chamberlain:
For some reason test_bad_interp only raises one warning on travis, but two locally - attempt to debug


calum-chamberlain:
@cjhopp looks like this fixed it - warnings was not warning more than once in separate tests, not sure why this used to work, but setting warning.simplefilter at the top has fixed it. If you think it's okay I'll merge once CI green-lights - everything looks like it's passing - I have left an unnecessary warning print in there, but not concerned by that.

calum-chamberlain:
Fixes fails, al fails due to time-out on OSX. Merging.

PR #106, opened 2017-07-19
Append streams when return_stream == True

calum-chamberlain:
This PR changes `eqcorrscan.core.match_filter.Tribe.client_detect` to return the whole stream used for the detection run if `return_stream` is set to True.  I think this is the expected behaviour, but it will lead to large amounts of memory being taken up if the users is not careful.


calum-chamberlain:
Tests failing due to timeouts - not associated, merging.

PR #107, opened 2017-07-22
Version 0.2.0

calum-chamberlain:
This PR constitutes the updates for version 0.2.0. See the change-log for details.

Once this merges the following needs to happen:

1. Update DOI;
2. pypi release;
3. conda release;
4. Email to users group.


calum-chamberlain:
@cjhopp 
All the changes have been reviewed to get into develop (or deemed minor, or too big to properly review...), so, the main point of the review here would be to say whether you think we are ready for another release?

I think we are - I will work on the openMP loop for PAN specifically in a separate branch and that will probably constitute a minor (0.2.1) release later - but there is a lot in here, quite a few bug catches and some major API changes (e.g. objects), and underlying changes (e.g. your and @d-chambers work to subspace, and mine and @d-chambers work on match-filter internals).  I think there is definitely more to go (which is fine), but there is enough benefit to users to suggest they update...?

calum-chamberlain:
I was thinking of enforcing an overlap in Party detection methods so that nothing is missed. Should probably include that...

calum-chamberlain:
It was going to be 1.0.0, but I'm not convinced I am ready for the big leap to 1.0.0 status yet! I don't think it is quite stable enough - I still have a lot more ideas to implement, so I think 0.2.0 is the way to go - still signifies large changes from 0.1.x.

Anyway - I'm going to try and get two more things in:
1 [ ] Allow overlap with value in `Tribe.detect` and `Tribe.client_detect` methods;
2 [ ] Add note in docs about overlaps/check that note is there
3 [ ] Make `utils.clustering.distance_matrix` use the faster multi-comparison methods.

I will try and get those done in PR's to develop in the next couple of days and then update this PR and merge.

Issue #126, opened 2017-08-07
Lag-calc method for multiple templates sometimes gives the wrong answer


calum-chamberlain:
This issue relates to `eqcorrscan.core.lag_calc.lag_calc`. When using multiple templates, LagCalcErrors are raised, when for the same templates and stream, running templates individually does not raise such an error.  The following (large - need to generate a smaller tests-case) results in the error:
```python
from eqcorrscan.core.match_filter import Tribe
from eqcorrscan.utils.catalog_utils import filter_picks
from eqcorrscan.utils.plotting import (pretty_template_plot,
                                       detection_multiplot)
from obspy.clients.fdsn import Client
from obspy import UTCDateTime

client = Client('NCEDC')
t1 = UTCDateTime(2004, 9, 28)
t2 = t1 + 86400
catalog = client.get_events(
    starttime=t1, endtime=t2, minmagnitude=2.5, minlatitude=35.7,
    maxlatitude=36.1, minlongitude=-120.6, maxlongitude=-120.2,
    includearrivals=True)
catalog = filter_picks(catalog, top_n_picks=5)
tribe = Tribe().construct(
     method='from_client', catalog=catalog, client_id='NCEDC', lowcut=4.0,
     highcut=15.0,  samp_rate=40.0, filt_order=4, length=6.0, prepick=0.1,
     swin='all', process_len=21600)
repicked_catalog = party.lag_calc(
     stream, pre_processed=False, shift_len=0.2, min_cc=0.4)
```
However, running the following loop raise no error:
```python
from obspy.core.event import Catalog
repicked_catalog = Catalog()
for family in party:              
    print(family)
    repicked_catalog += family.lag_calc(stream, pre_processed=False, shift_len=0.2, min_cc=0.4)
```

This issue possibly relates to `eqcorrscan.core.lag_calc._prepare_data`.

To do:
 - [x] Prepare a smaller broken example
 - [x] Debug!

calum-chamberlain:
This is being worked on in PR #128

Issue #122, opened 2017-08-07
Tribe and Party should allow slicing by template name.


calum-chamberlain:
It would be helpful for Tribe and Party objects to allow extraction of members by template name.

calum-chamberlain:
Closed by #133

PR #128, opened 2017-08-07
Lag-calc method for multiple templates sometimes gives the wrong answer.

calum-chamberlain:
This PR relates to issue #126 regarding `eqcorrscan.core.lag_calc.lag_calc`. When using multiple templates, LagCalcErrors are raised, when for the same templates and stream, running templates individually does not raise such an error.  The following (large - need to generate a smaller tests-case) results in the error:
```python
from eqcorrscan.utils.catalog_utils import filter_picks
from eqcorrscan.utils.plotting import (pretty_template_plot,
                                       detection_multiplot)
from obspy import UTCDateTime

client = Client('NCEDC')
t1 = UTCDateTime(2004, 9, 28)
t2 = t1 + 86400
catalog = client.get_events(
    starttime=t1, endtime=t2, minmagnitude=2.5, minlatitude=35.7,
    maxlatitude=36.1, minlongitude=-120.6, maxlongitude=-120.2,
    includearrivals=True)
catalog = filter_picks(catalog, top_n_picks=5)
tribe = Tribe().construct(
     method='from_client', catalog=catalog, client_id='NCEDC', lowcut=4.0,
     highcut=15.0,  samp_rate=40.0, filt_order=4, length=6.0, prepick=0.1,
     swin='all', process_len=21600)
repicked_catalog = party.lag_calc(
     stream, pre_processed=False, shift_len=0.2, min_cc=0.4)
```
However, running the following loop raise no error:
```python
from obspy.core.event import Catalog
repicked_catalog = Catalog()
for family in party:              
    print(family)
    repicked_catalog += family.lag_calc(stream, pre_processed=False, shift_len=0.2, min_cc=0.4)
```

This issue possibly relates to `eqcorrscan.core.lag_calc._prepare_data`.

To do:
 - [x] Prepare a smaller broken example
 - [x] Debug!

When merged this should prompt a new patch release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 * Change internals of *eqcorrscan.core.lag_calc._prepare_data* to use a
   dictionary for delays, and to work correctly! Issues arose from not checking
   for masked data properly and not checking length properly.
+* Fix bug in match_filter.match_filter when checking for equal length traces,
+  length count was one sample too short.
 
 ## 0.2.4
 * Increase test coverage (edge-cases) in template_gen;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 ## Current
 * Fix bug with \_group_process
 * Force NumPy version
+* Add tests for lag-calc issue with preparing data
+* Change internals of *eqcorrscan.core.lag_calc._prepare_data* to use a
+  dictionary for delays, and to work correctly! Issues arose from not checking
+  for masked data properly and not checking length properly.
 
 ## 0.2.4
 * Increase test coverage (edge-cases) in template_gen;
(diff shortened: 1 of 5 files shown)


calum-chamberlain:
State of play as of first commit:
- a few changes made to cope with little bugs in lag-calc 
- syntax changes
- start a test that is known to fail locally - this is a mammoth test and a smaller, similar example should be found before proceeding with debugging.

calum-chamberlain:
The issue appears to relate to using merged data with overlaps - they were not handled correctly before in match-filter (merge should use method=1), and then a check for length was not sufficiently performed in _prepare_data, nor was the mask check sufficient in _prepare_data.  A small test has been prepared for merged data.

calum-chamberlain:
Not sure about patch release - there are other things that I would like to include before a patch release, but if anyone is having a similar issue they should raise it in the issues page.

calum-chamberlain:
@cjhopp This fixes a few minor bugs that came to light when playing with different processing-lengths for the example in the manuscript - I really want to merge this and push a patch release before submitting the paper, so providing the ci passes, could you do a quick review (of this and #132 and #122) please?

PR #132, opened 2017-08-12
Demean the image to normalise time-domain cross-correlation

### LINKED ISSUE #131: Time-domain normalisation
Time-domain cross-correlation in EQcorrscan was not originally designed to be used for long time-series with variable means, and as such does not implement normalisation correctly.  If we are to expose time-domain methods more readily (as we would in #119), then this function needs to be adjusted.

In general I think this should be corrected anyway so that users do not get unexpected results.

calum-chamberlain:
This PR close #131


calum-chamberlain:
@cjhopp This passes locally - all I'm doing is removing the mean of every moving window, which gets me the correct result (and compares favourably to the frequency-domain x-correlations). If the tests pass on the CI, can I merge this?

calum-chamberlain:
Appveyor timeout for single-test not affected, not re-running.

PR #133, opened 2017-08-15
Add support for indexing by template-name

### LINKED ISSUE #122: Tribe and Party should allow slicing by template name.

calum-chamberlain:
This PR adds support for extracting a Template from a Tribe by the name of the template, and similarly, extracting a Family from a Party by the name of the template.

Closes #122

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Current
 * Fix bug with \_group_process
 * Force NumPy version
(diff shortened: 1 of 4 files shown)


calum-chamberlain:
Does not work on py 2.7 - need to check for all string types (unicode etc).

PR #138, opened 2017-08-17
Adding OpenMP support in multi_corr.c

chrisdjscott:
Not ready to merge yet but we can keep track of changes here.


eqcorrscan/lib/multi_corr.c
```diff
@@ -316,13 +373,180 @@ int normxcorr_time(float *template, int template_len, float *image, int image_le
...
+        /* get the id of this thread */
+        tid = omp_get_thread_num();
+        #endif
+
+        /* initialise memory to zero */
+        memset(template_ext[tid], 0, (long) fft_len * n_templates * sizeof(double));
+        memset(image_ext[tid], 0, (long) fft_len * sizeof(double));
+
+        /* call the routine */
+        r = normxcorr_fftw_main(&templates[(long) n_templates * template_len * i], template_len,
```
chrisdjscott: I had to cast to `long` when calculating the array offsets for larger datasets, to avoid integer overflow (weird behaviour). Might be worth passing `int` arguments in as `long` from Python?

calum-chamberlain: Curious, but we can definitely do that if we need to. I know that @d-chambers is working on changes to the `eqcorrscan.utils.correlate` - changing to long in there might result in a conflict when we merge but it should be straightforward to manage.

chrisdjscott:
Initial benchmarking results comparing the OpenMP version with the Python multiprocessing version (both call the same underlying C function: `normxcorr_fftw_main`). These benchmarks were with the real data and 100 templates.

results_100templates

Memory usage (top graph):
* Peak memory usage increases with core count and is slightly lower for OpenMP
* The number of cores is limited by memory and the number of channels (9 in this case)

Run time (bottom graph):
* OpenMP much faster on 1 core but Python catches up as the number of cores increases
* This is because more code is parallelised in the Python version (see below)
* OpenMP slightly faster at 9 (= num_channels) cores

results_py_vs_c

* In the OpenMP version there is some serial stuff done in Python before and after the call to C (see `utils.correlate.fftw_multi_normxcorr`)
* This gets run on each thread (i.e. in parallel) in the multiprocessing version (see `utils.correlate.fftw_normxcorr`)
* As we increase cores the serial Python code becomes a bottleneck in the openmp version

Next I plan to look at:

* Adding threading in the main C routine that computes the cross correlation (`normxcorr_fftw_main`)
  - should allow us to utilise more cores
  - increasing cores here should not increase memory usage
  - will need to do some benchmarking to see if it is worthwhile
* Look at the serial Python code in the OpenMP version to see if it can be optimised or is worth running in parallel

eqcorrscan/lib/multi_corr.c
chrisdjscott: We could probably leave it as it is for now, as it only seems to be an issue when computing the offsets, and think about changing later so that it doesn't conflict with the other changes.

chrisdjscott:
I have modified the Python function to move the parts that were taking significant time into C, so it can run in parallel (C should be faster too). I profiled that function using [line_profiler](https://pypi.python.org/pypi/line_profiler) first, see below if interested. The "% Time" column is important. Note only 57.5% in the C function (some lines with very low "% Time" omitted).

```
Function: fftw_multi_normxcorr at line 342

Line        Hits         Time  Per Hit   % Time  Line Contents
==============================================================
...
   399         1            8      8.0      0.0      cccs = np.empty((n_channels, n_templates, image_len - template_len + 1),
   400         1      6440302 6440302.0    8.3             np.float32).flatten(order='C')
   401         1            6      6.0      0.0      ret = utilslib.multi_normxcorr_fftw(
   402         1            2      2.0      0.0          template_array, n_templates, template_len, n_channels, stream_array,
   403         1     44357571 44357571.0   57.5          image_len, cccs, fft_len)
...
   408        10           21      2.1      0.0      for j in range(n_channels):
   409       909         2002      2.2      0.0          for i in range(n_templates):
   410       900         2129      2.4      0.0              if not used_chans[j][i]:
   411       593      4266116   7194.1      5.5                  cccs[j][i] = np.zeros(image_len - template_len + 1)
   412         1      2517897 2517897.0     3.3      cccs[np.isnan(cccs)] = 0.0
   413         1      8582607 8582607.0     11.1      if np.any(np.abs(cccs) > 1.01):
   414                                                   print('Normalisation error in C code')
   415                                                   print(cccs.max())
   416                                                   print(cccs.min())
   417                                                   raise MemoryError()
   418         1      2552231 2552231.0      3.3      cccs[cccs > 1.0] = 1.0
   419         1      2518577 2518577.0      3.3      cccs[cccs < -1.0] = -1.0
   420        10           40      4.0      0.0      for j, seed_id in enumerate(seed_ids):
   421       909         4753      5.2      0.0          for i in range(len(pad_array[seed_id])):
   422       900         3602      4.0      0.0              cccs[j][i] = np.append(
   423       900         4176      4.6      0.0                  cccs[j][i],
   424       900      5841140   6490.2      7.6                  np.zeros(pad_array[seed_id][i]))[pad_array[seed_id][i]:]
   425         1           13     13.0      0.0      cccs = cccs.reshape(n_channels, n_templates, image_len - template_len + 1)
   426         1            2      2.0      0.0      return cccs, used_chans
```

We can remove the `flatten` on line 400 since the array is guaranteed to be `C_CONTIGUOUS` - when you access the underlying data in C it will be implicitly flattened. The array in this line was very big and `flatten` creates a full copy of the array, so that's probably why it was taking so long. The other time consuming lines can be moved to C. 

After doing this we get the following updated graph (OMP2 is the new version).

results_100templates_omp2

- Memory usage is similar
- Run time is significantly lower with OMP2
  * Multiprocessing with 9 threads takes 89.7 seconds
  * Original OpenMP with 9 threads takes 80.4 seconds
  * New OpenMP with 9 threads takes 51.2 seconds (1.8x faster than multiprocessing and 1.6x faster than the original OpenMP version)

I still want to look at threading in the main C routine and maybe run some bigger tests on this version.

calum-chamberlain:
Just had a thought: In the python function that calls the python function that calls the c-function (convoluted) within correlate.py (multi_normxcorr.py?) we are summing the results of the c-function as soon as we get them back.  It would be more memory efficient to compute the sums in the c-function, as long as it doesn't impact parallel computation. Do you think that would be feasible? We don't need the cross-correlations for individual channels, just the sum of all channels for each template.
What do you think @chrisdjscott ?

chrisdjscott:
Yes, sounds like it might make sense to do this in the C code too @calum-chamberlain - I'll have a look into it.

chrisdjscott:
@calum-chamberlain - I moved the sum into C and we got a small boost in performance - 47.0 seconds down from 51.2s (OpenMP) and 89.7s (multiprocessing), so about 1.9x faster than multiprocessing now. Memory usage doesn't change much as I think the full array is still needed for the calculation.

I'll try merging the changes from develop into here to get rid of the conflict (and see if any change in performance). Then I will look to see if threading in `normxcorr_fftw_main` will help. Does that sound OK?

calum-chamberlain:
Nice, okay - I was thinking that the summing could go within `normxcorr_fftw_main` - to replace the full array of cross-correlation (rather than assigning each value, each value could be added to the previous one), which might have reduced memory cost.  I guess that might not work calling in parallel and potentially trying to sum multiple times to the same bit of memory?

chrisdjscott:
We would have to use atomic operations to do that I think @calum-chamberlain, which could slow things down. I'll have a look to see how easy it will be to implement as it would be good to get an idea of the performance cost. It looks like about 12GB memory is used in that big array (in the test case you sent me with 9 channels and 100 templates), while only ~1.3GB needed after the sums so could save a reasonable amount.

calum-chamberlain:
That is a concept far beyond my grasp of computing...

chrisdjscott:
If you mark a specific operation (in this case we will be doing a `+=` on the ncc array) as `omp atomic` it means that only one thread can execute that line at any one time, so you may have some threads doing nothing while they wait for another thread to finish executing the line. It looks like I can try it out without too much work; I'll do that now.

If that doesn't work well the other option is that each thread could have its own copy of the summed array, so you don't have to worry about other threads accessing it at the same time, and then merge them together at the end. This would save memory from what we're currently doing if the number of threads is less than the number of channels.

calum-chamberlain:
That could be good - I guess you would end up doing most of the waiting for locks early in the summing process? Assuming that the add operations took similar time for each step through the array (which sounds like a poor assumption to me). If it isn't too difficult it would be good to see what the performance impact is, both on time and memory - if we can reduce the memory a lot, then we can potentially run a lot more data at the same time...

chrisdjscott:
Yes that could well be the case.

Maybe not going to be so easy to test. There is a whole bunch of stuff being done, such as checking for nan, checking for normalisation errors and applying pad_array, that would also need to be moved into `normxcorr_fftw_main`. I need to have a closer look but I think it could make sense for all that stuff to be done in `normxcorr_fftw_main`, as then it would also benefit the multiprocessing version (which also ends up calling that function). 

I think the first thing I will do is get up to date with the develop branch and check everything is still working, as its probably easier to fix conflicts now than later.

chrisdjscott:
@calum-chamberlain I made the change you suggested to do the sum in place in the main C routine. Some results from Pan below. The latest is actually fastest, so the benefit of moving the checks/post processing into the same loop as the calculation (so getting rid of another loop) outweighs any downside to having the atomic operation. Also, memory usage is down by over 10GB.

I will try and merge from develop now (I may need some advice on how to update my benchmark script to use the new interface).

| Version  | Run time (s) | Peak memory (GB) |
|---------|--------------|------------------|
| Multiproc (original)       | 89.7 ± 0.1   | 135.6 |
| Multiproc (post proc in C) | 81.8 ± 0.1   | 123.8 |
| OpenMP (first go)          | 80.4 ± 0.1   | 113.2 |
| OpenMP (post proc in C)    | 51.2 ± 0.1   | 113.2 |
| OpenMP (accumulate in C)   | 47.0 ± 0.1   | 113.1 |
| OpenMP (sum in place)      | 44.7 ± 0.1   | 102.3 |

calum-chamberlain:
Ace, that looks good! And yup if you remind me of the benchmark script (maybe a gist?) then I can have a look and see what if anything needs to change.

chrisdjscott:
Here is one of the scripts I'm using @calum-chamberlain 

https://gist.github.com/chrisdjscott/777c8a534df5739c7e8157337c720d79

Will the calls to `multichannel_normxcorr` have to change?

calum-chamberlain:
Those calls will have to change, I'm don't think the old multi-process method is exposed at the moment though (see issue #151), so possibly now isn't the time to merge? If you have done, that is fine, but the old multiprocessing version won't be available until later today likely.

calum-chamberlain:
Also, the new way to call the functions is:
```python
multichannel_normxcorr = get_stream_xcorr(xcorr_func, concurrency)
[cccsums, no_chans, chans] = multichannel_normxcorr(
        templates=templates, stream=stream, cores=cores)
```
Where `xcorr_func` is either a string or a callable, and concurrency is a string or `None`.
Available strings for `xcorr_func` are: `time_domain`, `numpy`, `fftw`
Available strings for concurrency are: `multithread`, `multiprocess`, `concurrent`

Access to the openMP loop you have written should be through `xcorr_func='fftw', concurrency='concurrent'`.
Access to the old, multiprocessing loop should be through `xcorr_func='fftw', concurrency='multiprocess'`, but I will need to play for a bit to get that available.

chrisdjscott:
OK, sounds good. I was only using the old multiprocessing one as a reference to check the results from OpenMP and multiprocessing were the same when running the real dataset, so could be replaced with something else I guess. Probably makes sense to wait for #151.

calum-chamberlain:
So I think #145 now has everything we need in it, and it passes tests locally, I'm going to wait for tests to run on the ci then merge into develop, then we should update this branch.

chrisdjscott:
Cool, once that's in develop I'll do the merge into here.

calum-chamberlain:
@chrisdjscott merged, go for it, sorry it took so long - was getting some annoying test crashes running threaded tests in parallel - have now separated them out so that threaded tests are called in serial.

calum-chamberlain:
Let me know if you have iassue with the merge conflicts, happy to take care of those two.  Sorry for changing multi_corr.c :-1:

chrisdjscott:
No worries, I'll work on the merge later today and let you know if I have any problems.

eqcorrscan/utils/correlate.py
```diff
@@ -582,18 +463,23 @@ def fftw_normxcorr(templates, stream, pads, threaded=False, *args, **kwargs):
         np.ctypeslib.ndpointer(dtype=np.intc,
                                flags=native_str('C_CONTIGUOUS'))]
     restype = ctypes.c_int
-    if threaded and False:  # Forcing non-threaded for now...
+
+    # TODO: convert threaded routine to do post processing in C too - forcing non-threaded for now...
```
calum-chamberlain: I think your stuff should be calling the multi_normxcorr_fftw function - these two just call the inner, single channel normxcorr_fftw(_threaded) functions.  These are here to cope with python looping over channels (either using multiprocessing, multithreading or in serial).  The concurrent option should direct to multi_normxcorr_fftw.

chrisdjscott: It's a bit confusing now. The OpenMP one calls `multi_normxcorr_fftw` which then calls `normxcorr_fftw_main` at each iteration of the loop. The multiprocessing/threading/serial version calls `normxcorr_fftw` when threading is set to `False` - this ends up calling `normxcorr_fftw_main` too. However the threaded version calls `normxcorr_fftw_threaded`.

When I made the change to do the sum in place I had to move the post processing code from both python functions into `normxcorr_fftw_main` in C.  I need to replicate this for `normxcorr_fftw_threaded` too otherwise the post processing won't get done in that case - this shouldn't take long.

calum-chamberlain: Ah, okay, cool, sorry, I didn't see that!

chrisdjscott:
Looks like the merge was successful. All the tests are passing on Pan, the multiprocessing and OpenMP versions give the same result for the real dataset and performance is basically the same (I like the new interface too). 

How do you want to proceed @calum-chamberlain? I think you said that at some point you would run some tests with a bigger dataset.

I would still like to look at threading in the main routine (as well as and instead of the outer loop) as memory usage would not increase with the number of threads there and may mean more data can be processed at once. Not clear what performance would be like yet though. Maybe worth doing this in a separate branch though, so we're not trying to do too much here?

eqcorrscan/utils/correlate.py
```diff
@@ -650,37 +654,30 @@ def fftw_multi_normxcorr(template_array, stream_array, pad_array, seed_ids):
...
-                         image_len - template_len + 1))
-    for j in range(n_channels):
-        for i in range(n_templates):
-            if not used_chans[j][i]:
-                cccs[j][i] = np.zeros(image_len - template_len + 1)
-    cccs[np.isnan(cccs)] = 0.0
-    if np.any(np.abs(cccs) > 1.01):
-        print('Normalisation error in C code')
+    elif ret > 0:
+        print('Error in C dode (possible normalisation error)')
```
calum-chamberlain: Typo - dode for code

calum-chamberlain:
That all looks really great - I will try and run some tests on PAN using my fuller Kaikoura dataset and see how far I can push it this week if I may?
I think it would be good to see if threading internally can be done and if it makes a difference to performance, but you are probably right, it's probably worth putting in another pull request.

chrisdjscott:
Yes sounds good about running some bigger tests on Pan. Let me know if you run into any problems.

chrisdjscott:
Ah, looks like there may be a problem with the OS X build, due to no `isnanf` - I'll have a look tomorrow.

chrisdjscott:
Fixing that `isnanf` warning hasn't fixed the seg fault. I'll try and reproduce on my macbook (no luck so far) and run a memory debugger on it to see if that throws anything out.

chrisdjscott:
I can reproduce the crash on my mac using gcc-4.9 (same as on travis), it doesn't happen with gcc 7.2.

chrisdjscott:
There is something strange going on here. I get the segmentation fault on my mac with gcc-4.9 from homebrew. However it works fine with gcc-4.8, gcc-5, gcc-6 and gcc-7 (all from homebrew). It could be there is something wrong with gcc-4.9, given that it works with other compilers and platforms. Also, it looks like the develop branch test failed with the same error so don't think it is something introduced here. Any thoughts @calum-chamberlain?

calum-chamberlain:
Agreed, that is super strange - and yes, it does look like develop failed on that issue: previously, before it merged, it didn't fail on that, but did timeout. Looks like I ignored something somewhat important.
I'm going to check that travis works with gcc-6, if it does we can carry on with that and I can try and work out what is going on with gcc-4.9 on my Mac later (don't have it in the office).

calum-chamberlain:
Yup, so travis runs fine with gcc-6 - how about you update from develop again (which will just change the .travis.yml file)., and I will investigate further gcc-4.9 later.

I'm hoping to run some bigger tests tomorrow afternoon - lecture prep is taking over my mind at the moment. Apologies for the delay.

calum-chamberlain:
Looks good - just running some bulk tests on PAN, then I will merge.

calum-chamberlain:
Larger test (with >2000 templates through 60 days of data - split into chunks) ran happily with no errors due to this.
Merging!
👍

chrisdjscott:
Awesome, I'll start looking at threading within the main function.

Issue #112, opened 2017-07-29
Could not load shared library


d-chambers:
Running Ubuntu 16.04 I followed the instructions listed [here](http://eqcorrscan.readthedocs.io/en/latest/intro.html#installation-updated-for-version-0-2-0), including creating a conda virtual env and installing fftw. 

I then cloned the current repo and tried to run the test suite but I get many failing tests with the following errors:

'Could not load shared library "libutils.cpython-36m-x86_64-linux-gnu.so".\n\n /media/data/Gits/EQcorrscan/eqcorrscan/lib/libutils.cpython-36m-x86_64-linux-gnu.so: undefined symbol: GOMP_parallel_sections'

Next, I tried deactivating the conda env and installing from source (master) using:
```bash
python setup.py install 
```
The installation seems to have worked but re-running the tests the same problem arises.

calum-chamberlain:
Rubbing the tests requires you to run

python setup.py build_ext --inplace

Calum Chamberlain
Postdoctoral Fellow, Seismology
Victoria University of Wellington
Sent from mobile.

________________________________

d-chambers:
Ok I ran the build_ext command, then tried to run the tests and get the same error. This time I installed eqcorrcan with the develop flag, not sure if that makes a difference.

calum-chamberlain:
Does the file in question exist? If not then the build has gone wrong. If yes then something has gone wrong in the linking, if this is the case you should paste here the GCC lines.

Calum Chamberlain
Postdoctoral Fellow, Seismology
Victoria University of Wellington
Sent from mobile.

________________________________

calum-chamberlain:
Ah, on reading more closely, it looks like you don't have the openmp libraries installed? I had thought this was standard in gcc, more info to follow (and will update docs once we get this working).

p.s. eqcorrscan is now on condo-forge, and comes with fftw and openmp when installing from there, but you can't run the tests from there...

calum-chamberlain:
It looks like a linking error to the openMP libraries, but I don't know why it would be (tests on 16.04 run fine with standard setups).  Can I check that gcc is the compiler that gets called by setup.py build_ext? It would usually be gcc unless you built your Python in an interesting way (although the current method also works for intel compilers).

d-chambers:
I am just running anaconda with python version 3.6, shouldn't be anything unusual about it. 

[...]

PR #139, opened 2017-08-26
Correlate refactor

d-chambers:
close issue # 119

Hey @calum-chamberlain and @cjhopp , I think this is able to merge. I would appreciate it if you could give it a good look-over to make sure I didn't do anything dumb in handling merge conflicts. We will see how the CI runs go...

### DIFF
--- a/eqcorrscan/__init__.py
+++ b/eqcorrscan/__init__.py
@@ -7,19 +7,13 @@
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
-import sys
 import importlib
+import sys
 import warnings
-
-# "shortcut" imports
-from eqcorrscan.utils.correlate import register_array_xcorr
-
 
 __all__ = ['core', 'utils', 'tutorials']
 
-
 __version__ = '0.2.5'
-
 
 # Cope with changes to name-space to remove most of the camel-case
 _import_map = {
@@ -39,6 +33,7 @@ class EQcorrscanRestructureAndLoad(object):
     """
     Path finder and module loader for transitioning
     """
+
     def find_module(self, fullname, path=None):
         # Compatibility with namespace paths.
         if hasattr(path, "_path"):
@@ -91,4 +86,5 @@ sys.meta_path.append(EQcorrscanRestructureAndLoad())
 
 if __name__ == '__main__':
     import doctest
+
     doctest.testmod(exclude_empty=True)
--- a/eqcorrscan/__init__.py
+++ b/eqcorrscan/__init__.py
@@ -11,6 +11,10 @@ import sys
 import importlib
 import warnings
 
+# "shortcut" imports
+from eqcorrscan.utils.correlate import register_array_xcorr
+
+
 __all__ = ['core', 'utils', 'tutorials']
 
 
--- a/eqcorrscan/core/match_filter.py
+++ b/eqcorrscan/core/match_filter.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import ast
+import contextlib
 import copy
 import getpass
 import glob
@@ -45,6 +46,16 @@ from eqcorrscan.utils.correlate import get_array_xcorr, get_stream_xcorr
 from eqcorrscan.utils.findpeaks import find_peaks2_short, decluster
 from eqcorrscan.utils.plotting import cumulative_detections
 from eqcorrscan.utils.pre_processing import dayproc, shortproc
+
+
+@contextlib.contextmanager
+def temporary_directory():
+    """ make a temporary directory, yeild its name, cleanup on exit """
+    dir_name = tempfile.mkdtemp()
+    yield dir_name
+    if os.path.exists(dir_name):
+        shutil.rmtree(dir_name)
+
 
 
 def _spike_test(stream, percent=0.99, multiplier=1e6):
@@ -597,26 +608,23 @@ class Party(object):
             if os.path.exists(filename):
                 raise IOError('Will not overwrite existing file: %s'
                               % filename)
-            os.makedirs(filename)
-            Tribe([f.template for f in self.families]).write(
-                filename=filename, compress=False)
-            all_cat = Catalog()
-            for family in self.families:
-                all_cat += family.catalog
-            if not len(all_cat) == 0:
-                all_cat.write(filename + os.sep + 'catalog.xml',
-                              format='QUAKEML')
-            for i, family in enumerate(self.families):
-                print('Writing family %i' % i)
-                name = family.template.name + '_detections.csv'
-                name_to_write = join(filename, name)
-                _write_family(family=family, filename=name_to_write)
-            with tarfile.open(filename + '.tgz', "w:gz") as tar:
-                tar.add(filename, arcname=os.path.basename(filename))
-            if os.path.isdir(filename):
-                shutil.rmtree(filename)
-            elif os.path.isfile(filename):
-                os.remove(filename)
+            # os.makedirs(filename)
+            with temporary_directory() as temp_dir:
+                Tribe([f.template for f in self.families]).write(
+                    filename=temp_dir, compress=False)
+                all_cat = Catalog()
+                for family in self.families:
+                    all_cat += family.catalog
+                if not len(all_cat) == 0:
+                    all_cat.write(join(temp_dir, 'catalog.xml'),
+                                  format='QUAKEML')
+                for i, family in enumerate(self.families):
+                    print('Writing family %i' % i)
+                    name = family.template.name + '_detections.csv'
+                    name_to_write = join(temp_dir, name)
+                    _write_family(family=family, filename=name_to_write)
+                with tarfile.open(filename + '.tgz', "w:gz") as tar:
+                    tar.add(temp_dir, arcname=os.path.basename(filename))
         else:
             warnings.warn('Writing only the catalog component, metadata '
                           'will not be preserved')
--- a/eqcorrscan/core/match_filter.py
+++ b/eqcorrscan/core/match_filter.py
@@ -1694,7 +1694,8 @@ class Template(object):
 
     def detect(self, stream, threshold, threshold_type, trig_int, plotvar,
                pre_processed=False, daylong=False, parallel_process=True,
-               ignore_length=False, overlap="calculate", debug=0):
+               xcorr_func=None, concurrency=None, ignore_length=False,
+               overlap="calculate", debug=0):
         """
         Detect using a single template within a continuous stream.
 
@@ -1731,6 +1732,21 @@ class Template(object):
             over other methods.
         :type parallel_process: bool
         :param parallel_process:
+        :type xcorr_func: str or callable
+        :param xcorr_func:
+            A str of a registered xcorr function or a callable for implementing
+            a custom xcorr function. For more details see
+            :func:`eqcorrscan.utils.correlate.register_array_xcorr`.
+        :type concurrency: str
+            The type of concurrency to apply to the xcorr function. Options are
+            'multithread', 'multiprocess', 'concurrent'. For more details see
+            :func:`eqcorrscan.utils.correlate.get_stream_xcorr`.
+        'multithread', 'multiprocess', 'concurrent'. For more details see
+        :func:`eqcorrscan.utils.correlate.get_stream_xcorr`
+        :param concurrency:
+            The type of concurrency to apply to the xcorr function. Options are
+            'multithread', 'multiprocess', 'concurrent'. For more details see
+            :func:`eqcorrscan.utils.correlate.get_stream_xcorr`.
         :type ignore_length: bool
         :param ignore_length:
             If using daylong=True, then dayproc will try check that the data
@@ -1812,8 +1828,9 @@ class Template(object):
             templates=[self], stream=stream.copy(), threshold=threshold,
             threshold_type=threshold_type, trig_int=trig_int,
             plotvar=plotvar, pre_processed=pre_processed, daylong=daylong,
-            parallel_process=parallel_process,
-            ignore_length=ignore_length, overlap=overlap, debug=debug)
+            parallel_process=parallel_process, xcorr_func=xcorr_func,
+            concurrency=concurrency, ignore_length=ignore_length,
+            overlap=overlap, debug=debug)
         return party[0]
 
     def construct(self, method, name, lowcut, highcut, samp_rate, filt_order,
@@ -2236,8 +2253,9 @@ class Tribe(object):
         return tribes
 
     def detect(self, stream, threshold, threshold_type, trig_int, plotvar,
-               daylong=False, parallel_process=True, ignore_length=False,
-               group_size=None, overlap="calculate", debug=0):
+               daylong=False, parallel_process=True, xcorr_func=None,
+               concurrency=None, ignore_length=False, group_size=None,
+               overlap="calculate", debug=0):
         """
         Detect using a Tribe of templates within a continuous stream.
 
@@ -2267,6 +2285,16 @@ class Tribe(object):
             over other methods.
         :type parallel_process: bool
         :param parallel_process:
+        :type xcorr_func: str or callable
+        :param xcorr_func:
+            A str of a registered xcorr function or a callable for implementing
+            a custom xcorr function. For more information see:
+            :func:`eqcorrscan.utils.correlate.register_array_xcorr`
+        :type concurrency: str
+        :param concurrency:
+            The type of concurrency to apply to the xcorr function. Options are
+            'multithread', 'multiprocess', 'concurrent'. For more details see
+            :func:`eqcorrscan.utils.correlate.get_stream_xcorr`
         :type ignore_length: bool
         :param ignore_length:
             If using daylong=True, then dayproc will try check that the data
@@ -2383,6 +2411,7 @@ class Tribe(object):
                 threshold_type=threshold_type, trig_int=trig_int,
                 plotvar=plotvar, group_size=group_size, pre_processed=False,
                 daylong=daylong, parallel_process=parallel_process,
+                xcorr_func=xcorr_func, concurrency=concurrency,
                 ignore_length=ignore_length, overlap=overlap, debug=debug)
             party += group_party
         for family in party:
@@ -2970,8 +2999,8 @@ def _test_event_similarity(event_1, event_2, verbose=False):
 
 def _group_detect(templates, stream, threshold, threshold_type, trig_int,
                   plotvar, group_size=None, pre_processed=False, daylong=False,
-                  parallel_process=True, ignore_length=False,
-                  overlap="calculate", debug=0):
+                  parallel_process=True, xcorr_func=None, concurrency=None,
+                  ignore_length=False, overlap="calculate", debug=0):
     """
     Pre-process and compute detections for a group of templates.
 
@@ -3017,6 +3046,15 @@ def _group_detect(templates, stream, threshold, threshold_type, trig_int,
         over other methods.
     :type parallel_process: bool
     :param parallel_process:
+    :type xcorr_func: str or callable
+    :param xcorr_func:
+        A str of a registered xcorr function or a callable for implementing
+        a custom xcorr function. For more details see:
+        :func:`eqcorrscan.utils.correlate.register_array_xcorr`
+    :type concurrency: str
+        The type of concurrency to apply to the xcorr function. Options are
+        'multithread', 'multiprocess', 'concurrent'. For more details see
+        :func:`eqcorrscan.utils.correlate.get_stream_xcorr`
     :type ignore_length: bool
     :param ignore_length:
         If using daylong=True, then dayproc will try check that the data
@@ -3097,6 +3135,7 @@ def _group_detect(templates, stream, threshold, threshold_type, trig_int,
             detections += match_filter(
                 template_names=[t.name for t in template_group],
                 template_list=[t.st for t in template_group], st=st_chunk,
+                xcorr_func=xcorr_func, concurrency=concurrency,
                 threshold=threshold, threshold_type=threshold_type,
                 trig_int=trig_int, plotvar=plotvar, debug=debug, cores=ncores)
             for template in template_group:
@@ -3573,7 +3612,7 @@ def match_filter(template_names, template_list, st, threshold,
     :type xcorr_func: str or callable
     :param xcorr_func:
         A str of a registered xcorr function or a callable for implementing
-        a custom xcorr function. For more information check out:
+        a custom xcorr function. For more information see:
         :func:`eqcorrscan.utils.correlate.register_array_xcorr`
     :type concurrency: str
     :param concurrency:
@@ -3699,6 +3738,30 @@ def match_filter(template_names, template_list, st, threshold,
         0.1 seconds early. We are working on a solution that will involve
         saving templates alongside associated metadata.
 
+    .. Note:: xcorr_func Custom xcorr functions can be used according to the following:
+    
+    .. rubric:: Example
+    >>> import obspy  
+    >>> from eqcorrscan.utils.correlate import time_multi_normxcorr
+    >>> # define a custom xcorr function
+    >>> def custom_normxcorr(templates, stream, pads, *args, **kwargs):
+    ...     # Just to keep example short call other xcorr function 
+    ...     # in practice you would define your own function here
+    ...     print('calling custom xcorr function')
+    ...     return time_multi_normxcorr(templates, stream, pads, *args, **kwargs) 
+    >>> # generate some toy templates and stream
+    >>> random = np.random.RandomState(42)
+    >>> template = obspy.read()
+    >>> stream = obspy.read()
+    >>> for num, tr in enumerate(stream):  # iter stream and embed templates
+    ...     data = tr.data
+    ...     tr.data = random.randn(6000) * 5
+    ...     tr.data[100: 100 + len(data)] = data
+    >>> # call match_filter ane ensure the custom function is used 
+    >>> match_filter(['1'], [template], stream, .5, 'absolute', 1, False, 
+    ...              xcorr_func=custom_normxcorr)  # doctest: +SKIP
+
+    calling custom xcorr function ... # doctest: +SKIP
     """
     _spike_test(st)
     import matplotlib
--- a/eqcorrscan/core/match_filter.py
+++ b/eqcorrscan/core/match_filter.py
@@ -18,32 +18,33 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import numpy as np
-
-import warnings
 import ast
-import os
-import time
 import copy
 import getpass
-import re
-import tarfile
-import shutil
-import tempfile
 import glob
-
-from multiprocessing import cpu_count
+import os
+import re
+import shutil
+import tarfile
+import tempfile
+import time
+import warnings
 from collections import Counter
-from obspy import Trace, Catalog, UTCDateTime, Stream, read, read_events
-from obspy.core.event import Event, Pick, CreationInfo, ResourceIdentifier
-from obspy.core.event import Comment, WaveformStreamID
+from multiprocessing import cpu_count
+from os.path import join
 
+import numpy as np
+from obspy import Trace, Catalog, UTCDateTime, Stream, read, read_events
+from obspy.core.event import Comment, WaveformStreamID
+from obspy.core.event import Event, Pick, CreationInfo, ResourceIdentifier
+
+from eqcorrscan.core import template_gen
+from eqcorrscan.core.lag_calc import lag_calc
+from eqcorrscan.utils.catalog_utils import _get_origin
+from eqcorrscan.utils.correlate import get_array_xcorr, get_stream_xcorr
 from eqcorrscan.utils.findpeaks import find_peaks2_short, decluster
 from eqcorrscan.utils.plotting import cumulative_detections
 from eqcorrscan.utils.pre_processing import dayproc, shortproc
-from eqcorrscan.utils.catalog_utils import _get_origin
-from eqcorrscan.core import template_gen
-from eqcorrscan.core.lag_calc import lag_calc
 
 
 def _spike_test(stream, percent=0.99, multiplier=1e6):
@@ -75,6 +76,7 @@ class MatchFilterError(Exception):
     """
     Default error for match-filter errors.
     """
+
     def __init__(self, value):
         """
         Raise error.
@@ -111,6 +113,7 @@ class Party(object):
     """
     Container for multiple Family objects.
     """
+
     def __init__(self, families=None):
         """Instantiate the Party object."""
         self.families = []
@@ -511,7 +514,7 @@ class Party(object):
         detect_info = [(d[1], _total_microsec(d[0].datetime, min_det.datetime))
                        for d in detect_info]
         peaks_out, inds_out = decluster(
-            peaks=detect_info, trig_int=trig_int * 10**6, return_ind=True)
+            peaks=detect_info, trig_int=trig_int * 10 ** 6, return_ind=True)
         # Trig_int must be converted from seconds to micro-seconds
         declustered_detections = [all_detections[ind] for ind in inds_out]
         # Convert this list into families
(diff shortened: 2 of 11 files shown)


calum-chamberlain:
Grand - I won't be able to have a proper review of this today, but hopefully on Wednesday.

In the mean-time, could you fix up the pep8 things from stickler, looks like a lot of appveyor fails were FDSN issues (which is a near constant annoyance now), although this fail:
```python
eqcorrscan\tests\correlate_test.py:383: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
name = 'normxcorr'
    def gen_xcorr_func(name):
        """ return an xcorr function with desired name """
    
        def func(templates, stream, pads, *args, **kwargs):
            pass
    
>       func.__name__ = name
E       TypeError: __name__ must be set to a string object
eqcorrscan\tests\correlate_test.py:33: TypeError
```
Looks like a str/unicode py2/3 issue.

setup.py
```diff
@@ -257,7 +257,7 @@ def setup_package():
             'Development Status :: 4 - Beta',
             'Intended Audience :: Science/Research',
             'Topic :: Scientific/Engineering',
-            'License :: OSI Approved :: GNU Library or Lesser General Public ' +
+            'License :: OSI Approved :: GNU Library or Lesser General Public '
```
calum-chamberlain: Why the change here? Was the license not rendering as expected (e.g. with the addition of the 'License (LGPL)' line below?

eqcorrscan/__init__.py
```diff
@@ -11,6 +11,10 @@
 import importlib
 import warnings
 
+# "shortcut" imports
+from eqcorrscan.utils.correlate import register_array_xcorr
```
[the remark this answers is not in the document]
calum-chamberlain: Should be able to ignore this by changing to:
```python
from eqcorrscan.utils.correlate import register_array_xcorr  # noqa: F401
```

calum-chamberlain:
Regarding checking the docs, I can't get readthedocs to build your fork - I will have a look over them locally, but sometimes things go awry on RTD that don't get picked up locally, so we will need to:
-  [ ] check docs for develop once merged.

eqcorrscan/core/match_filter.py
```diff
@@ -605,12 +608,15 @@ def write(self, filename, format='tar'):
...
-                _write_family(
-                    family=family, filename=filename + os.sep +
-                    family.template.name + '_detections.csv')
+                name = family.template.name + '_detections.csv'
+                name_to_write = join(filename, name)
+                _write_family(family=family, filename=name_to_write)
             with tarfile.open(filename + '.tgz', "w:gz") as tar:
                 tar.add(filename, arcname=os.path.basename(filename))
-            shutil.rmtree(filename)
+            if os.path.isdir(filename):
```
calum-chamberlain: A test incorporating this if statement would be good - does look like it needs to be here - thanks!

eqcorrscan/core/match_filter.py
```diff
@@ -1528,24 +1535,20 @@ def __eq__(self, other):
                                       'calib']:
                             if tr.stats[trkey] != oth_tr.stats[trkey]:
                                 return False
-                elif isinstance(
-                        self.st, Stream) and not isinstance(other.st, Stream):
+                elif self_is_stream and not other_is_stream:
                     return False
-                elif not isinstance(
-                        self.st, Stream) and isinstance(other.st, Stream):
+                elif not self_is_stream and other_is_stream:
```
calum-chamberlain: Nice, much cleaner!

eqcorrscan/core/match_filter.py
```diff
@@ -3490,26 +3501,27 @@ def normxcorr2(template, image):
...
             templates=np.array([template]).astype(np.float32),
             stream=image.astype(np.float32), pads=[0], threaded=False)[0][0]
     ccc = ccc.reshape((1, len(ccc)))
     return ccc
 
 
 def match_filter(template_names, template_list, st, threshold,
-                 threshold_type, trig_int, plotvar, plotdir='.', cores=None,
+                 threshold_type, trig_int, plotvar, plotdir='.',
+                 xcorr_func=None, concurrency=None, cores=None,
```
calum-chamberlain: I think these extra options probably need to be included in:
- _group_detect (both below methods call this function)
- Template.detect
- Tribe.detect

setup.py
d-chambers: If I understand what you are doing here the plus just isn't needed, eg:
```python
a = 'version ' + '5.7'
b = 'version ' '5.7'
assert a == b
```
If you feel that it will make it more clear I can add it back, but it does knock the line length under 80.

calum-chamberlain: Ah, right you are - I was just following examples a while back I think that used that syntax.

d-chambers:
Looks like I am running into some problems with python versions less than 3.6, will dig into it and see if I can get the CI passing. Will also address any pep8 issues I missed.

eqcorrscan/tests/correlate_test.py
```diff
@@ -2,108 +2,399 @@
...
+        yield func.__name__
+        self.counter.pop(func.__name__)
+
+    # tests
+    def test_new_method_was_called(self, normxcorr_new_multithread):
+        """ ensure the new method was called """
+        assert self.counter[normxcorr_new_multithread]
 
 
 if __name__ == '__main__':
```
calum-chamberlain: Note to selves - because these are all pytest now, this `if __name__ == '__main__':` doesn't make sense/do anything - @d-chambers is there a py.test alternative?

d-chambers: From what I read [here](https://stackoverflow.com/questions/20992731/meaning-of-unittest-main-in-python-unittest-module), I don't think we need it. When calling `pytest` these tests will get auto-discovered and run. If you want to just run the tests in this file you can specify that when invoking pytest like so:
```bash
pytest eqcorrscan/tests/correlate_test.py
```
You can use pytest.main() as described [here](https://docs.pytest.org/en/latest/usage.html#calling-pytest-from-python-code) if you want to be able to run the tests from python rather than using the CLI. Is that not the same thing as `unittest.main()`?

eqcorrscan/utils/correlate.py
```diff
@@ -598,8 +598,10 @@ def fftw_multi_normxcorr(template_array, stream_array, pad_array, seed_ids):
     n_templates = template_array[seed_ids[0]].shape[0]
     image_len = stream_array[seed_ids[0]].shape[0]
     fft_len = next_fast_len(template_len + image_len - 1)
-    template_array = np.array(list(template_array.values())).flatten(order='C')
-    stream_array = np.array(list(stream_array.values())).flatten(order='C')
+    template_list = [template_array[x] for x in seed_ids]
+    stream_list = [stream_array[x] for x in seed_ids]
+    template_array = np.array(template_list).flatten(order='C')
+    stream_array = np.array(stream_list).flatten(order='C')
```
calum-chamberlain: This looks like it fixed those fails, nice catch - the failing tests look to me to all be network timeout related :-1: so I will rerun those until they don't crap out.
@chrisdjscott has started to move some of the reshaping into the C-func #138 to make this faster, and this may well clash with his thing, but I can work it out later on.  Looking at #138 though I am surprised that we haven't run into this issue there?

eqcorrscan/__init__.py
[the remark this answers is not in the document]
d-chambers: Hey @calum-chamberlain, how do you feel about imports here? Should I get rid of this or do you plan to have something accessible from the eqcorrscan namespace?

eqcorrscan/utils/correlate.py
chrisdjscott: @calum-chamberlain I should probably make this change in my branch too - it should be easy to merge later as only a couple of lines.

eqcorrscan/__init__.py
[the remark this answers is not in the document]
calum-chamberlain: I haven't done so far, but am now getting annoyed with having to type the whole chain to get what I want. I think we should work out what functions we want to be able to import directly from the eqcorrscan namespace though before going too deep into this. Can we leave this out for now and have a separate PR later?

eqcorrscan/utils/correlate.py
calum-chamberlain: Sounds good @chrisdjscott. I'm still not sure why this didn't emerge earlier in the testing from updating the internal correlation routines and am a little concerned. @d-chambers do you know if this fail happened thanks to a change in tests, or was it just from changing `correlate.py`?

eqcorrscan/core/match_filter.py
d-chambers: I think I just had to put this in to make the tests past; somehow I got in a state where they were failing because it was trying to delete a directory that didn't exist or was a file instead. Looking closer at the code, it does seem strange this can happen given line 600 should make the directory. I am not really sure what is going on here but I will try to look into it more.

eqcorrscan/utils/correlate.py
d-chambers: I have no idea but there clearly was an order dependence on un-ordered dictionaries. I have a feeling we would have to dive very deep into the C-python implementation to understand how it didn't affect the code before, but it is possible that it might have caused some subtle bugs that didn't fail the tests, who knows?

eqcorrscan/core/match_filter.py
d-chambers: Ok, should the parallel_process option be removed in favor of the concurrency argument? Maybe it should just be depreciated? If so do we have an official way of depreciating things?

eqcorrscan/tests/correlate_test.py
calum-chamberlain: So pytest.main doesn't do the same as unittest.main - I think we should just remove the whole `if __name__ == __main__` statement in here - calling via pytest is cleaner anyways.

eqcorrscan/utils/correlate.py
calum-chamberlain: Agreed on the dependence on the order - I hadn't even considered it because it never presented itself as an issue. I wonder if the order was preserved inside the function before, which has now been lost by genertaing the dictionaries in another funcion (which btw is IMHO much cleaner and better 👍 )? - either way this is the *right* way to do this and I should have done it originally. 

I'm just debating whether I patch master with this? I'm finding it hard to find a justification for why it used to work, which is pushing me towards patching. Open to opinions @cjhopp and @d-chambers ?

eqcorrscan/core/match_filter.py
calum-chamberlain: Agreed, there should be a directory - this is one instance that using the `tempfile.TemporaryDirectory` context manager would be really handy, but isn't available in py2.7 AFAIK. 
It really shouldn't ever be a file eh... Weird, I don't know how this could have happened, but if you need help let me know.

eqcorrscan/tests/correlate_test.py
```diff
@@ -2,108 +2,399 @@
...
 import time
+import unittest
+from collections import defaultdict
+from functools import wraps
 
+import eqcorrscan.utils.correlate
+import eqcorrscan.utils.correlate as corr
+import numpy as np
+import pytest
+from eqcorrscan import register_array_xcorr
```
calum-chamberlain: With the loss of the direct import this needs to be reverted back to:
```python
from eqcorrscan.utills.correlate import register_array_xcorr
```

eqcorrscan/tests/correlate_test.py
```diff
@@ -2,108 +2,399 @@
...
+
+import copy
+import itertools
 import time
+import unittest
+from collections import defaultdict
+from functools import wraps
 
+import eqcorrscan.utils.correlate
+import eqcorrscan.utils.correlate as corr
```
calum-chamberlain: This looks like a double import unless I am missing something?

eqcorrscan/tests/correlate_test.py
d-chambers: right you are. That is what I get for assuming a small change wont break everything :)

eqcorrscan/utils/correlate.py
d-chambers: Patching is not a bad idea unless you feel we are close to another version release.

eqcorrscan/core/match_filter.py
calum-chamberlain: So `parallel_process` is just supposed to determine whether the data should be pre-processed in parallel, it *should* be separate to the actual matched-filtering, so, we should be able to leave that, and just add in the `concurrency` and `xcorr_func` kwargs?

eqcorrscan/utils/correlate.py
calum-chamberlain: If I'm going to patch it then it would be another release - it would be nice to include this in the next release, it's feeling close to me, what do you guys think? @cjhopp @d-chambers 
If it is close (next week or so) then I'd be tempted to wait, mostly because I can't reproduce the issue with the old/current code...

eqcorrscan/tests/correlate_test.py
```diff
@@ -2,112 +2,395 @@
...
+    """ these tests ensure the various implementations of normxcorr return
+    approximately the same answers """
+
+    # tests
+    def test_single_channel_similar(self, array_ccs):
+        """ ensure each of the correlation methods return similar answers
+        given the same input data """
+        cc_list = list(array_ccs.values())
+        for cc1, cc2 in itertools.combinations(cc_list, 2):
+            assert np.allclose(cc1, cc2, atol=0.05)
```
calum-chamberlain: Can we push this atol lower? With properly normalised time-domain cross-correlations they should all be *more* similar now?

eqcorrscan/tests/correlate_test.py
```diff
@@ -2,112 +2,395 @@
...
+@pytest.fixture(scope='module')
+def stream_cc_dict(stream_cc_output_dict):
+    """ return just the cc arrays fro mthe stream_cc functions """
+    return {name: result[0] for name, result in stream_cc_output_dict.items()}
+
+
+# ----------------------------------- tests
+
+
+class TestArrayCorrelateFunctions:
```
calum-chamberlain: It may just be me not following the pytest syntax, but it's not clear that we are still testing every method (and every type of concurrency) are equal?

eqcorrscan/utils/clustering.py
```diff
@@ -13,24 +13,25 @@
...
 import os
-
+import warnings
 from multiprocessing import Pool, cpu_count
-from scipy.spatial.distance import squareform
-from scipy.cluster.hierarchy import linkage, dendrogram, fcluster
-from obspy.signal.cross_correlation import xcorr
+
+import matplotlib.pyplot as plt
+import numpy as np
```
calum-chamberlain: Is there any reason for moving the numpy import lower? I was having them at the top to follow ObsPy, but didn't have a better reason.

eqcorrscan/utils/correlate.py
```diff
@@ -24,18 +24,290 @@
...
+
+
+# ------------------ Context manager for switching out default
+class _Context:
+    """ class for permanently or temporarily changing items in a dict """
+
+    def __init__(self, cache, value_to_switch):
+        """
+        :param cache:
+        :param value_to_switch:
```
calum-chamberlain: Can we have some info for this doc-string?

eqcorrscan/utils/correlate.py
```diff
@@ -24,18 +24,290 @@
...
+        """
+        self.cache = cache
+        self.value_to_switch = value_to_switch
+        self.previous_value = None
+
+    def __call__(self, new_value, *args, **kwargs):
+        """ # TODO change docs if this ever becomes general use
```
calum-chamberlain: How do you see this becoming general use, do you mean if we expose this to the users? Are we exposing it and will they use it?

eqcorrscan/utils/correlate.py
```diff
@@ -24,18 +24,290 @@
...
 
-def numpy_normxcorr(templates, stream, pads):
+    see also :func:`eqcorrscan.utils.correlate.get_stream_xcorr`
     """
-    Compute the normalized cross-correlation of multiple templates with data.
+    func = _get_registerd_func(name_or_func)
+    return func.array_xcorr
+
+
+# def array_xcorr(templates, stream, pads, *args, **kwargs):
```
calum-chamberlain: Did we decide we didn't need this section? If so, should we remove it?

eqcorrscan/doc/submodules/utils.correlate.rst
```diff
@@ -17,5 +17,10 @@ correlate
        multichannel_normxcorr
        numpy_normxcorr
        time_multi_normxcorr
+       get_array_xcorr
+       get_stream_xcorr
+       register_array_xcorr
+       stream_norm_xcorr
+
```
calum-chamberlain: I wonder if this file would be a useful place to introduce the notion of swappable correlation functions, and a little example? It would also be good to have a short example of using match_filter with a different correlation function.

calum-chamberlain: Although that example should go in the match_filter docstring, I will comment there also.

eqcorrscan/core/match_filter.py
```diff
@@ -3550,6 +3570,16 @@ def match_filter(template_names, template_list, st, threshold,
...
+    :type xcorr_func: str or callable
+    :param xcorr_func:
+        A str of a registered xcorr function or a callable for implementing
+        a custom xcorr function. For more information check out:
+        :func:`eqcorrscan.utils.correlate.register_array_xcorr`
+    :type concurrency: str
+    :param concurrency:
+        The type of concurrency to apply to the xcorr function. Options are
+        'multithread', 'multiprocess', 'concurrent'. For more details see
+        :func:`eqcorrscan.utils.correlate.get_stream_xcorr`
```
calum-chamberlain: Can you add a short example of using a different xcorr_func to the default here please?

calum-chamberlain:
This is excellent, thank you massively for this, not only does this make the code better, but it will make updates with faster/better code, and code that is application appropriate simple and effective.
The main things that I would like to see from my review are:
- [ ] An example of changing the xcorr_func in the match_filter doc-string (probably just a small dummy data example, the main thing I want to see is how to get the xcorr_func before calling - on that note, you could just do an example in a correlate function doc-string and refer to that, up to you);
- [ ] `concurrency` and `xcorr_func` arguments for `_group_detect`, `Tribe.detect` and `Template.detect`.

The other couple of things (the commented out block, `__init__` docstring and your `# TODO` comment) are things that shouldn't get in the way, but I am curious about the `# TODO`.

eqcorrscan/tests/correlate_test.py
d-chambers: Ya, it is more concise now but does require a bit of knowledge about pytest fixtures. If you look at the fixture `array_ccs`  it is creating a dictionary. The key is the name of the method and the value is the result of calling the function with the same array_template and array_stream. 

Then `test_single_channel_similar` iterates over all combinations of the values (outputs from the different correlation methods) and makes sure the results are similar. 

I am not sure how you are running the tests but I find myself commenting out the adopts line in pytest.ini, adding `import pdb; pdb.set_trace()` to tests I am interested in looking closer at (for example in the `test_single_channel_similar` function), then running pytest by selecting the node of interest, eg `pytest eqcorrscan/tests/correlate_test.py::TeatArrayCorrelateFunction -s -x --pdb`. This will drop you into a debug prompt to poke around with what is going on in the tests. the s flag will disable output suppression, the x flag tells the tests to stop after one failure, and the pdb flag will drip you into a pdb console whenever a test fails. 

I have also found the pdb++ package very useful. Food for thought.

eqcorrscan/utils/clustering.py
d-chambers: I use pycharm as an IDE. After writing some code I usually apply use the code reformat function to clean up all the nit-picky pep-8 stuff. As a consequence, it sorts the imports into the following 3 categories, separated by a blank line, outlined by pep8's section on imports:

    standard library imports
    related third party imports
    local application/library specific imports

since numpy is not part of the standard library it gets put into the second category and alphabetized. I can put it back but this is the most pep8 compliant.

eqcorrscan/utils/correlate.py
d-chambers: Sure. It should be noted this is a private class, so I don't really expect it to be initialized apart from the instance directly after the class definition, but it is good to document its behavior for other developers.

eqcorrscan/utils/correlate.py
d-chambers: If we want to have something else that can be toggled based on context, maybe number of cores? My note is just to remove the specific mention of normxcorr if we start using it for other things too. Until/if that happens, it is better to leave the doc string the way it as in order to spoof the _Context instance assigned to the name "set_xcorr" into looking like a function

eqcorrscan/tests/correlate_test.py
calum-chamberlain: Cool, thanks for the notes - will have to start playing with this style of testing soon - maybe I'll convert some of the other tests slowly to pytest style.

When I'm checking individual tests I usually run:
```python
python <test-file.py> <TestClass>.<test_name>
```
Which is why the `if __name__ == '__main__'` bit was in the tests.

Also, haven't played with pdb, but have seen it mentioned variously, should play.
Thanks for the explanation!

eqcorrscan/utils/clustering.py
calum-chamberlain: Sweet, no, I'm not that picky about the imports, and if this is more pep8 friendly then it seems like a good way to go - I use pycharm too, should probably go through the whole codebase at some point and run the code reformat (lots of things in pycharm I don't know about!).

eqcorrscan/utils/correlate.py
calum-chamberlain: Agreed on private-ness, more that I don't immediately understand what these args are, just a brief note on the the arguments are would help me out :)

eqcorrscan/doc/submodules/utils.correlate.rst
d-chambers: Ok, I may need a bit of help with this one. In past projects I have always used jupyter notebooks as the main form of documentation, how would I add the example here? Just with markdown syntax after the sphinx stuff?

eqcorrscan/core/match_filter.py
d-chambers: Ok, does it go directly under the :param: in the docstring or at the end in an examples tag? With numpy style docs it would probably be put at the end under Examples section, not sure how you want me to do it here.

calum-chamberlain: Yup, you are correct, I hit the wrong line - example should go at the end after all the arguments and returns.  See [here](https://github.com/eqcorrscan/EQcorrscan/blob/master/eqcorrscan/utils/mag_calc.py#L85) for an example of an example in EQcorrscan.

eqcorrscan/doc/submodules/utils.correlate.rst
calum-chamberlain: Yup, markdown syntax is the way to do - you should check how it renders though by making the docs:
From within the EQcorrscan/eqcorrscan/docs directory run `make html`, then open the EQcorrscan/eqcorrscan/docs/html/index.html file in the browser and navigate to the page you want just to check that it looks like useful Python.

eqcorrscan/tests/correlate_test.py
d-chambers: When I lower the atol the tests fail. Maybe this branch doesn't have the commit from https://github.com/eqcorrscan/EQcorrscan/pull/121? I was able to find the commits in the git log though... I am a bit confused on this one.

calum-chamberlain: How low did you go? Is it obvious which correlation doesn't match the others, or do they all fail?

d-chambers: It was time-domain that was failing. Just had to delete the build directory and re-compile (I should have learned my lesson from https://github.com/eqcorrscan/EQcorrscan/issues/112)

d-chambers: test now pass with atol = .00001, do you want it lower or is that sufficient?

calum-chamberlain: That looks like a good level - I have also made an issue to remind me to change how build works (#141), although I'm not sure how to do it, but it would make avoiding these issues much simpler, and I think you suggested it in #112 and I forgot to actually do it?

eqcorrscan/doc/submodules/utils.correlate.rst
d-chambers: Ok, just out of curiosity is there a way to run the python in the markdown to make sure the docs stay fresh? To me, the appeal of jupyter notebooks is they can be run (even as part of the tests or separate CI pipe) to make sure nothing is broken before they are rendered to html.

calum-chamberlain:
Having the doctests fixture at the bottom will find and run all examples in the docstrings whenever tests are run.

eqcorrscan/doc/submodules/utils.correlate.rst
calum-chamberlain: As long as you have:
```python
if __name__ == '__main__':
    import doctest
    doctest.main()
```
at the bottom of the file (which most/all files should have that are not solely test files themselves), then purest will pick up the tests every time it runs, and you can run the file using pytest or python to run the doctests. Doctests will find and run all examples in the docs.

d-chambers: This applies for the .rst files in the doc directory as well?

calum-chamberlain: No, those aren't looked in. How I get those to run is a little convoluted, but I have done it for the tutorials:
- write a script and put it in the tutorials directory
- include that file in the .rst (with a line: `.. literalinclude:: ../../tutorials/<test_script>.py` in the rst)
- call that file from the tests/tutorial_tests.py file to run a test on it.

calum-chamberlain: Although, it looks like I could do it: https://docs.pytest.org/en/latest/doctest.html

calum-chamberlain: Note to self here: #142 
If you want to write your examples in rst let me know and I can prioritise getting that running in the next couple of days so you can rebase and run things.

eqcorrscan/doc/submodules/utils.correlate.rst
```diff
@@ -14,13 +14,101 @@ correlate
...
+    stream = obspy.read()
+    for num, tr in enumerate(stream):  # iter stream and embed templates
+        data = tr.data
+        tr.data = random.randn(6000) * 5
+        tr.data[100: 100 + len(data)] = data
+
+
+    # do correlation using numpy rather than fftw
+    match_filter(['1'], [template], stream, .5, 'absolute', 1, False,
+                 xcorr_func='numpy)
```
calum-chamberlain: close string around 'numpy'

eqcorrscan/doc/submodules/utils.correlate.rst
```diff
@@ -14,13 +14,101 @@ correlate
...
+    with set_xcorr(custom_normxcorr):
+        match_filter(['1'], [template], stream, .5, 'absolute', 1, False)
+        #  prints "calling custom xcorr function"
+
+    # you can also permanently change the xcorr function (until your python
+    # kernel is restarted) by calling set_xcorr
+    set_xcorr(custom_normxcorr)
+    match_filter(['1'], [template], stream, .5, 'absolute', 1, False)
+    # prints "calling custom xcorr function
+    set_xcorr.revert()  # change it back
```
calum-chamberlain: Great, thanks, this looks good!

d-chambers:
Ok, take a look. I am running into two issues I need your help with:

1. in utils.correlate.rst the link to `match_filter` works, but the link to the detect methods does not. 
2. I couldn't get doctest ellipses to work with the example in the `match_filter` docstring. It keeps failing despite having a matching substring so I had to just skip it. 

I will fix the sticker stuff in a minute

calum-chamberlain:
One final things while you catch those pep8 niggles: can you make a note in CHANGES.md to show that you have updated the internals of correlate.py and added concurrency and xcorr_func arguments to the upper API levels to allow interchangeable correlation functions, or something to that effect/more accurate.
After that I think we should merge if tests pass?

calum-chamberlain:
I will take a look at your two points now, sorry, didn't see them before above comment.

d-chambers:
What tag should this go under in the change log?

d-chambers:
I guess just start a new one for 0.2.6?

calum-chamberlain:
Can you make a new heading called:
`#Current`

calum-chamberlain:
@d-chambers can you point me to the line in utils.correlate.rst where the link isn't working? Can't immediately see it.

calum-chamberlain:
1. Change those two links to `eqcorrscan.core.match_filter.Tribe.detect` and `eqcorrscan.core.match_filter.Template.detect` - you were missing the `match_filter` bit.
2. Rewriting as below seems to work on my mac, maybe try this (note I also included imports that are not needed for doctest, but make the example work stand-alone for anyone trying it, which I prefer)?
```python
    >>> import obspy
    >>> import numpy as np
    >>> from eqcorrscan.core.match_filter import match_filter
    >>> from eqcorrscan.utils.correlate import time_multi_normxcorr
    >>> # define a custom xcorr function
    >>> def custom_normxcorr(templates, stream, pads, *args, **kwargs):
    ...     # Just to keep example short call other xcorr function 
    ...     # in practice you would define your own function here
    ...     print('calling custom xcorr function')
    ...     return time_multi_normxcorr(templates, stream, pads, *args, **kwargs) 
    >>> # generate some toy templates and stream
    >>> random = np.random.RandomState(42)
    >>> template = obspy.read()
    >>> stream = obspy.read()
    >>> for num, tr in enumerate(stream):  # iter stream and embed templates
    ...     data = tr.data
    ...     tr.data = random.randn(6000) * 5
    ...     tr.data[100: 100 + len(data)] = data
    >>> # call match_filter ane ensure the custom function is used 
    >>> detections = match_filter(
    ...     ['1'], [template], stream, .5, 'absolute', 1, False, 
    ...      xcorr_func=custom_normxcorr)  # doctest:+ELLIPSIS
    calling custom xcorr function...
```

- Failing tests at Travis look like network timeouts so far, so I wouldn't worry about them

d-chambers:
Thanks both issues seemed to be fixed now. On the rst docs I knew I was missing something obvious but it didn't jump out at me. I guess the # doctest:+ELLIPSIS line has to be on the line that writes to std.out? I just figured it could go anywhere on the example to enable the feature

calum-chamberlain:
It looks like there is an issue on travis using xdist and the threaded routines - I can reproduce on my desktop, and can stop it being an issue by including `--boxed` in pytest.ini [addopts].
This runs each test in a subprocess, and seems to avoid issues with globals/locals not being accessed properly.

d-chambers:
Weird, a NameError on a built-in function like print, how does that even happen?

calum-chamberlain:
It is really strange eh? I do not understand it well at all, I found [this](https://github.com/micropython/micropython/pull/2918) on my travels, which is somewhat relevant.
At a guess, I think it may be down to how xdist spawns threads (maybe before spawning CPython instances), which are then not honoured by the threading threads? But I really haven't the foggiest why built-in functions wouldn't be there. Very strange...
Doesn't seem to run as fast with `--boxed`, which might be an issue.

calum-chamberlain:
Using `--boxed` seems to be giving me an error with a brightness-test now, that a directory isn't found.

d-chambers:
I see whats going on. The doctest is adding the `custom_normxcorr` to correlate.XCORR_FUNCS dict so it is getting scoped up and tested in conjunction with the the built-in xcorr functions. It must have some weird name-spacing going on between the pytest namespaces and doctest namespaces. I should be able to fix this by limiting which functions are tested. Give me a minute...

calum-chamberlain:
Unsure why dependency-ci isn't reporting, or why appveyor/branch isn't reporting (this may be because this is from your fork actually).  But, I'm happy with the PR passes on both appveyor and travis, and the codecov status.
If you are happy then I will merge?

d-chambers:
Ya, looks good to me.

PR #144, opened 2017-09-06
WIP: Update utils.correlate.rst

### LINKED ISSUE #140: Correlation equation should be given
There are multiple ways to compute cross-correlations - we should be explicit in the docs of match_filter and utils.correlate about which equation we use (it is the same as Matlab normxcorr2 in the time-domain, and the frequency domain does the same thing too, but in the frequency domain...).

- [ ] docs for `utils.correlate`
- [ ] docs in `core.match_filter`
- [ ] docs in `core.lag_calc`

calum-chamberlain:
Update the docs for correlate to include the equation used - closes #140 

Also need to build these docs on readthedocs and check that they work properly.

### DIFF
--- a/eqcorrscan/doc/submodules/utils.correlate.rst
+++ b/eqcorrscan/doc/submodules/utils.correlate.rst
@@ -24,7 +24,7 @@ correlate
     .. comment to end block
 
 Selecting a correlation function
---------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


calum-chamberlain:
I had to write a dummy function for readthedocs for `next_fast_len` - I could have mocked it probably, but this way any documentation that has to run that uses correlate functions can use then (rather than getting stuck calling a `Mock` object).
Docs online look good to me @cjhopp - I'll let the tests run just in case, but I haven't changes any real code, so this should be fine to merge?

PR #145, opened 2017-09-06
Speed-up the time-domain a bit by running it in parallel

calum-chamberlain:
This PR changes the c routine for time-domain cross-correlations.  It now pre-computes the running mean (which cannot be computed in parallel), and the computes the normalised cross-correlation in parallel using openMP.

Need to check how this works with different methods of concurrency...


calum-chamberlain:
Tested on a large machine - this is not a safe option for `concurrency="multiprocess"`.

I should write a threaded and non-threaded version and tie the non-threaded version to the multiprocess option, and this to the multithread option (but avoiding the Python threading).  This is also fastest only for large thread-counts.

calum-chamberlain:
Tests pass on ci - now separating into two testing steps, a parallel set of tests and a non-parallel.  The non-parallel ones include threaded routines.
This is done using the `@pytest.mark.serial` decorator.

PR #148, opened 2017-09-06
Make sure an excess of cores isn't allocated

calum-chamberlain:
Testing on an 80 CPU machine, the tribe methods do not allocated the number of cores for parallel processing when using the construct method, which results in more threads being spawned than necessary for the data, and memory allocation errors.

This PR asserts that when `num_cores` is not set, the maximum that will be allocated is the number of channels in the stream.


calum-chamberlain:
appveyor fails are single timeouts that are unrelated.

calum-chamberlain:
Passes on travis, but OSX was taking forever so cancelled, will merge after #145

PR #158, opened 2017-09-13
Doctest examples

### LINKED ISSUE #142: Make pytest run examples in .rst files in eqcorrscan/docs
This should be possible by setting the `--doctest-glob` flag for pytest.

Doing this will be a useful update for the docs to make sure little things in the tutorials get caught (the main tutorials are run as tests on their own, but the docs are not).

calum-chamberlain:
Fixes #142 

This pull request reformats the appropriate doc-strings so that py.test will find them and run them.  I have set it up so that these get run on circleci, some are slow (network) tests, so running them there should result in less slow-don for other ci services.

I will need to build the docs on readthedocs and check that everything renders properly:
- [x] tutorials
- [x] utils.correlate

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,7 @@ jobs:
           command: |
             . venv/bin/activate
             py.test -m "network" -n 2
+            py.test eqcorrscan/doc/tutorials/*.rst eqcorrscan/doc/submodules/*.rst --cov-append
 
       - run:
           name: Upload to codecov
--- a/eqcorrscan/doc/submodules/utils.correlate.rst
+++ b/eqcorrscan/doc/submodules/utils.correlate.rst
@@ -89,36 +89,35 @@ for example:
 
 .. code-block:: python
 
-    import obspy
-    from eqcorrscan.utils.correlate import numpy_normxcorr, set_xcorr
-    from eqcorrscan.core.match_filter import match_filter
+    >>> import obspy
+    >>> import numpy as np
+    >>> from eqcorrscan.utils.correlate import numpy_normxcorr, set_xcorr
+    >>> from eqcorrscan.core.match_filter import match_filter
 
 
-    # generate some toy templates and stream
-    random = np.random.RandomState(42)
-    template = obspy.read()
-    stream = obspy.read()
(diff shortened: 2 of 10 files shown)


calum-chamberlain:
Docs are [here](http://eqcorrscan.readthedocs.io/en/doctest-examples/) - They look good to me.

calum-chamberlain:
This did raise an issue though: one for @d-chambers 

Calling `set_xcorr` fails with the following error:
```python
TypeError('__repr__ returned non-string (type NoneType)',)
```

I could reproduce this on my machine outside of doctest too, but I'm not sure of the best way to fix it.  Suggestions?

calum-chamberlain:
@d-chambers I added a print string that should output which item in the dict has been changed and what to.  This passes tests now, and although I like the idea of it staying silent, I'm not sure how it would work.

`__repr__` is now:
```python
    def __repr__(self):
        """ this hides the fact _Context instance are returned after calls """
        name = self.cache[self.value_to_switch].__str__()
        if hasattr(self.cache[self.value_to_switch], '__name__'):
            name = self.cache[self.value_to_switch].__name__
        out_str = ("%s changed to %s" % (self.value_to_switch, name))
        return out_str
```

d-chambers:
hey @calum-chamberlain so what is raising the TypeError? When pytest runs the doc examples? 
Looking through the code it seems the biggest change is just to add ">>>" and "..." to the code snippets so the rst files look like doctests right? 

There really is no reason the function needs to stay silent, it just seemed cool for it to behave like a function and a context manager. No problems here.

calum-chamberlain:
Ace! Thanks, and yup, that was it for running doctests in the rst files - really simple.  

I'm deliberately separating those out again, along with the slow network tests (that had a tendency to time out), because they are pretty network dependent and slow. To do that I'm just pointing py.test to them rather than including them in the main test suite - so if you run those tests at home just using `python setup.py test` or `py.test` without arguments, it won't run those doc-tests.  

I couldn't think of a good way to keep them separate while having them in the main test suite - if i could mark them (with a pytest.mark) I would, but I don't see a way for the doctests.

PR #164, opened 2017-09-16
Use floats instead of doubles

### LINKED ISSUE #163: Test using floats rather than doubles for Template and Image and fftwf library.
Memory is always an issue. I'm pretty sure I tested this when I developed the C-code, but I didn't make notes on whether I needed doubles for the template and image - I know we need doubles for the moving mean, but not sure about the template and image.

Should be easy to re-code and we have the tests in place to quickly test the accuracy...

calum-chamberlain:
Closes #163 

This branch changes from using double precision for extended templates and images (and cross-correlations) to using single-precision floats. It also changes to using the fftwf functions and fftwf_complex numbers.

Tests pass locally on my OSX machine - need to compare memory usage...

### DIFF
--- a/eqcorrscan/lib/multi_corr.c
+++ b/eqcorrscan/lib/multi_corr.c
@@ -45,10 +45,12 @@ int normxcorr_fftw(float*, int, int, float*, int, float*, int, int*, int*);
 
 static inline int set_ncc(int t, int i, int template_len, int image_len, float value, int *used_chans, int *pad_array, float *ncc);
 
-int normxcorr_fftw_main(float*, int, int, float*, int, float*, int, double*, double*, double*,
-        fftw_complex*, fftw_complex*, fftw_complex*, fftw_plan, fftw_plan, fftw_plan, int*, int*);
+int normxcorr_fftw_main(float*, int, int, float*, int, float*, int, float*, float*, float*,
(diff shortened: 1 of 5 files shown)


calum-chamberlain:
Looks like tests are haps - I think when I was testing what precision I needed to use I ended up leaving in doubles for everything when we only need it for the variables included in the moving mean calculation...

What do we think @eqcorrscan/maintainer ?

chrisdjscott:
If the precision is acceptable this a great change. I just ran the benchmark on Pan that I've been using and it's nearly 3x faster than the original multiprocessing now and much less memory:

| Version  | Run time (s) | Peak memory (GB) |
|---------|--------------|------------------|
| Multiproc (original)       | 89.7 ± 0.1   | 135.6 |
| OpenMP (double)      | 44.7 ± 0.1   | 102.3 |
| OpenMP (float)      | 30.7 ± 0.1   | 53.1 |

calum-chamberlain:
I'm fairly sure the precision is acceptable - in previous versions I was using half-precision floats for the correlations before summing them anyway. This version seems to get the correct result to within our threshold limits anyway (tests pass), perfect correlations are out by < 0.00001, which IMHO is good enough for seismic data.

Thanks @chrisdjscott for running those tests - very gratifying to see a large memory drop!

PR #166, opened 2017-09-17
Pull for 0.2.7 patch release.

calum-chamberlain:


calum-chamberlain:
Tests failing because [obspy.org](obspy.org) is down.

Issue #159, opened 2017-09-13
Development request: speed-up findpeaks


calum-chamberlain:
A request was made outside of github to speed-up find-peaks.

I think this could be done, and could likely be done with a C-routine.  I will have a play and see if I can get it to work.

Alongside this, a suggestion of testing difference in detections between single-channel thresholds and stacked and summed thresholds was made.  I think this would be worth experimenting with in a branch too.

d-chambers:
If I get a few minutes this weekend (not 100% sure I will) I can take a look at this. 

I think you could do it all in numpy, getting rid of the python lists will go a long way. That would be worth trying before jumping into C. 

Can you ask the user to provide an example dataset? Even if we speed it up for the test cases it could be quite a different story for actual data.

calum-chamberlain:
You make an excellent point - if you do get the chance this weekend that would be excellent! Definitely keen to keep it in Python if we can.

I will get in touch for a test-dataset.

yifanyin:
Hi guys! Here comes the github outsider. The reason I suggested this is because I perform cross-correlation for a long period of time (several years) using few stations (2-5) so I end up running find_peak for lots of times. I tried to parallelize the find_peak in the old eqcorrscan version by the same way it parallelizes match_filter but discarded it since it upgraded to Python 3.x.

About the single-channel threshold. I intended to use it as a threshold in addition to the summed one. I got questioned that my detections are dominated by S-arrivals. Independent lower threshold might serve as a double-checker.

calum-chamberlain:
Do you think weighting the cross-correlation would help? You could weight template channels that contain only the P-wave more heavily than those containing only S-wave?
The reason I ask is that that is something that William Frank does with his correlations, the C-routines are on github [here](https://github.com/beridel/fast_matched_filter). These are just C-routines with Python and Matlab wrappers, so they don't have the front end that EQcorrscan does, but you can now define a custom xcorr function in EQcorrscan so you could install their package and test against that? It would require a bit of tweaking in EQcorrscan mind (to pass weights through), but if it is worth it it could be done.

A quick note if you do look at their correlations - they don't normalise in the same way that we do, so the correlations will be slightly different - for every step of the correlations they don't remove the mean of the continuous data.

d-chambers:
Hey @calum-chamberlain,
[...]

PR #167, opened 2017-09-20
WIP: Findpeaks optimisation

calum-chamberlain:
Not ready to merge yet, work in progress.

So far we have changed decluster to a more efficient NumPy implementation (thanks to @d-chambers), and compared to a c-implementation of decluster.  Speeds were similar, but we will go with the Python for simplicity and maintainability.

Still to do:
- [x] test for looping;
- [x] serial looping over multiple cccsum arrays;
- [x] parallel looping over multiple cccsum arrays;
~~- [ ] benchmarks.~~

This is work on the request in #159


calum-chamberlain:
Okay, so I have implemented a parallel loop and this is now called from matched-filter, which should speed things up a bit overall.

I also put the compiled version back in because testing the current decluster gave slower results than expected (about 10x slower than compiled). When I did that and tested them against one another, for the test dataset they both give the same results, but for a noisy randomised dataset they gave different results.  On closer inspection I saw that the compiled version was taking peaks happily, but the Python version missed some and took smaller adjacent peaks.  I will try and work up a test and put it up here.

d-chambers:
If the C version is more correct and much faster we can just use that; EQcorrscan already has C code that needs to be compiled so one more function doesn't really add a great deal of complexity right?

So do you need help adding tests that prove correctness, or do you still want to look at some benchmarking? If the C function is that much faster, and the python version is not even correct anyway, it may not be worth trying to fix the python version or coming up with benchmarks.

Nice use of the pool boy :wink:

calum-chamberlain:
Note, I may just be feeding the decluster function the wrong things. - @d-chambers do you want to gave a go at making it work? Otherwise I can tomorrow?

calum-chamberlain:
Fair enough - I'm fairly sure the C is correct, and reasonably fast. It is really simple C-code, doesn't require any weird libraries and doesn't care about system (ASFAIK), so yup, the C might be worth just going with rather than expending more effort.

Also, couldn't help myself with pool_boy, one of my new favourite bits of EQcorrscan. :small_airplane: 

I also moved `time_func` into utils.timer because I think it will be useful for more than just correlate.

I will just:
- [ ] go with the C version
- [ ] make a test or two more to test correctness
- [ ] add a more thorough test for finding all-peaks (I have left this option in, because I feel like I might want it for something else at some point)
- [ ] Document the limitation of find-peaks for closely spaced peaks that do not drop below the threshold between them.

I think that should do it for this one. @yifanyin is this the kind of thing you were expecting? After we merge this into develop we can create an experimental branch off develop for individual channel thresholding if you would like?

calum-chamberlain:
In saying that I think it is correct, now not so sure with the circleci fail for declustering detections.  Will check.

calum-chamberlain:
Just needed to use longs rather than ints for trig_int.

d-chambers:
I am having issues with my ubuntu machine (wont boot, black screen) this may take awhile to sort out. Will try to write some more tests when it is back up.

calum-chamberlain:
Oh no!

Looks like I have somehow introduced a memory leak in this latest one, so I would avoid running it until I fix that otherwise it will crash your machine again.

calum-chamberlain:
Looks like the memory leak doesn't happen on my laptop... curious.

yifanyin:
These codes are getting so advanced I can hardly keep up and find all the nobs I can turn ;)
I now tested eqcorrscan 0.2.7 on my machine and it works fine. However, on a 32-core machine, the multiprocess seems to freeze up and won't go into _find_peaks. Still trying to sort it out why. I'll try to test the new _find_peaks if someone can direct me to the right version!

calum-chamberlain:
@yifanyin, that sounds like it freezes in the cross correlations? If so, what are you setting for the xcorr_func, concurrency and cores arguments?

yifanyin:
@calum-chamberlain My setting: xcorr_func=None, concurrency="multiprocess", cores=30. Now I see where are the option of the xcorr_func haha. I'll also try alter them.

calum-chamberlain:
Try using concurrent as the concurrency argument. It will use the openmp parallel loop, which should be faster than multiprocess. Neither should use more cores than you have channels though, so if you have less than 30 channels I would recommend just sticking the cores argument to be the number of channels.

Calum Chamberlain
Sent from mobile.

calum-chamberlain:
Try using concurrent as the concurrency argument. It will use the openmp parallel loop, which should be faster than multiprocess. Neither should use more cores than you have channels though, so if you have less than 30 channels I would recommend just sticking the cores argument to be the number of channels.

**Although**: _pool_boy should take care of limiting the cores to be a maximum of the number of traces, so that argument shouldn't be an issue - you may just have to run shorter chunks of data if your memory is filling...?

d-chambers:
I re-factored the tests a bit to make it more modular and easier to add new test cases to TestPeakFindSpeeds. Let me know what you think. 

Would it be worthwhile to also allow for a concurrency option? Not sure how a thread pool would compare to a process pool. The tests are seg faulting after pulling the latest changes to the C routine so I can't really test it atm.

calum-chamberlain:
I shall see if I can reproduce and work out that segfault. 
I don't think a multi thread would be with it, but you are welcome to try if you want to?

d-chambers:
ok, I just needed to recompile, segfault is gone my bad

calum-chamberlain:
Those tests look good, took a couple of looks to get my head around the requests thing, but seems good to me.

~~I can't immediately reproduce the seg-fault, have you recompiled?~~

Above comment, seet, I think that would have been due to the change from ints to floats - I was having to use longs rather than straight ints for large gaps between things, but, in this version, the intervals between peaks are normalised by trig_int, which results in smaller gaps (smaller numbers to store in memory).  The main reason for doing this is because longs are not consistent between 32-Bit and 64-Bit machines, but Python always passes longs as int64 (on 32-Bit machines, long is int32...), so on 32-Bit machine, passing an array of longs doesn't work so well. This seemed like the simplest way around that.  It does slow down the Python part a bit, because an array of ints is converted to floats and divided by trig_int.

calum-chamberlain:
Okay, so, do we want to add more tests for correctitude (Friday words), or are we happy with this as is? And do you want to try a threading option?

Not sure why stickler hasn't run, but glad that all others are passing. I have tested install on all my machine (windows 32Bit, Ubuntu 64-Bit and OSX 64-Bit), but haven't tested on larger machines, but `pool_boy` should take care of CPU allocations so that it doesn't go crazy, so I think I'm happy on that front.

Also, I should think about using `pool_boy` for other multiprocessing things in EQcorrscan...

d-chambers:
I think it is good, we can add more tests as bugs crop up

calum-chamberlain:
Sweet, appveyor is confusing me. I'll fix those and be merge tomorrow.

I'm not planning on doing another release until 0.3.0 now so this won't make it out into the wild for a bit. Unless bugs are raised in the current release...

PR #168, opened 2017-09-29
Threading in normxcorr_fftw_main

chrisdjscott:
Work in progress.

Adding threading in here may give better performance/memory usage than on the outer loop (or maybe both can be used together).


chrisdjscott:
Some timings for the different sections of the routine (other parts were taking close to no time), before and after adding threading here (all on Pan Ivybridge nodes using `module load Python/3.6.0-gimkl-2017a`.

Initial timings (100 templates, no threads):

| Section | Time (s) | Percentage |
|---------|----------|------------|
| FFT of templates | 46.759269 | 39.3 |
| FFT of images | 0.335080 | 0.3 |
| dot product | 4.040679 | 3.4 |
| inverse FFT | 47.652992 | 40.1 |
| normalisation | 20.041162 | 16.9 |

After adding threads for FFTW calls, OpenMP for dot product, splitting the normalisation into two loops and linking against MKL instead of FFTW (running on 8 threads):

| Section | Time (s) | Percentage |
|---------|----------|------------|
| FFT of templates | 6.729761 | 32.9 |
| FFT of images | 0.073131 | 0.4 |
| dot product | 1.605530 | 7.9 |
| inverse FFT | 7.319455 | 35.8 |
| normalisation | 4.682571 | 22.9 |

Good speedup in each section.

chrisdjscott:
Benchmarking results for 100 templates. The best we got with parallelising the outer loop (after switching to single precision) was 30.7 seconds and 53.1 GB memory (using 9 threads).

time_100

* Using `module load Python/3.6.0-gimkl-2017a` is slightly faster than the `intel-2015a` module
* linking against mkl is better (`-lfftw3xf_gnu_pic -lmkl_rt` instead of the FFTW libraries)
* 8 threads: 23.8 seconds (~9.5 GB memory)
* 24 threads: 17.4 seconds (~9.5 GB)
* Faster than outer threading
* Memory usage lower

calum-chamberlain:
Wow.

chrisdjscott:
Tried using more templates too, now that memory usage is lower.

time_templates

memory_templates

chrisdjscott:
Looks like windows build failing due to no `collapse` in openmp loop - I'll run a benchmark without it to see if it makes any difference to performance but can probably just remove.

calum-chamberlain:
Wow.

calum-chamberlain:
circleci fails are just network issues as far as I can see, which is why those tests are seperate because they are vulnerable! Ignore them for the time being.

chrisdjscott:
Couple of things:

* how to decide whether to link with MKL or FFTW (at the moment I made it so you can uncomment some lines in setup.cfg to override the defaults from setup.py)
* could pass in the number of cores as an argument from python, to let the user choose (like it is done in the other functions)

calum-chamberlain:
- Is MKL always faster? If so, it might be worth having a check in setup.py to work out which libraries are available, and use the best. I'm not sure how to do that, but I can certainly have a look.
 - I do like being able to pass the number of cores, I have cases when using a shared resource and I don't want to use all of the cores so that others can.  I would advocate for this.  This also means that it is worth working out what the maximum number of useful cores is (to use when the default for `cores` is given), I'm not sure what that would be related to with this change, any idea?

chrisdjscott:
For the tests I've run so far MKL has always been faster. It might not be simple to detect in the setup.py but it's worth thinking about. For example, on Pan if you load the `Python-3.6.0-gimkl-2017a` module you have to link against `-lfftw3xf_gnu_pic -lmkl_rt` but if you load the `Python/3.5.0-intel-2015a` you link against `-lfftw3xf_intel_pic -lmkl_rt`, so slightly different. 

If MKL is installed then the `MKLROOT` environment variable should be set so that is one way of checking if MKL is there I suppose. I'll have a look into it a bit more.

chrisdjscott:
Looks like you only need to link against `-lmkl_rt` after all.

calum-chamberlain:
That's cool - I think we might be able to test for this on travis and appveyor MKL is available via conda, so we should be able to install them (providing travis and appveyor run intel chips) and have some kind of test that checks that the correct libraries are linked... Not sure how I would do that at the moment, and possibly doesn't require a full re-run of the test cases.

chrisdjscott:
Just had a look at the conda mkl package and we may need to use the intel channel - the default mkl does not include any header files (i.e. fftw3.h) but you can get them with `conda install -c intel mkl mkl-include`. I'll look in a bit more detail tomorrow.

chrisdjscott:
Actually it looks like mkl-include has been moved into the official anaconda channel recently (https://anaconda.org/anaconda/mkl-include) so it could be installed from there.

@calum-chamberlain do you want me to try adding a test for the MKL version on travis? One approach could be to add an extra env variable to the matrix - for the current tests set `MKL=0` and add a new test (e.g. just for the `PYTHON_VERSION=3.6` for now) with `MKL=1`. Then later on if `MKL==1` install mkl-include into the conda environment and it will get picked up during the build. Let me know what you think.

calum-chamberlain:
Hey @chrisdjscott, yes, if you can add that test that would be most excellent.  I have been playing with this branch locally, I think it is really desirable to be able to set the number of threads (either for inner and outer, or just for outer and calculate inner) from Python.  Have you test various ratios to try and work out what works best? Although "works best" probably means both which is the most memory efficient or the fastest, either would be helpful?

Thanks

chrisdjscott:
OK @calum-chamberlain, I'll add the test.

I can also add the option for setting the number of threads. I'll run some more benchmarks with different combinations of inner and outer to get the full picture first. For memory efficiency I think inner will always win as memory will always go up when you add more outer threads. Not so sure about performance so I'll run some more benchmarks.

chrisdjscott:
Strange, the travis OS X builds are failing due to not finding the `gcc-6` command although it should be installed with homebrew (maybe something to do with this line: `Homebrew must be run under Ruby 2.3! (RuntimeError)`).

chrisdjscott:
@calum-chamberlain I have run some more benchmarks, including rerunning the outer threading only cases using MKL instead of FFTW. Everything here is on an Ivybridge node on Pan, using `module load Python/3.6.0-gimkl-2017a` and linking against MKL.

results_0100

results_0200

results_1000

results_1990

* Some results missing for 3 and 9 outer threads at higher numbers of templates as they ran out of memory.
* It's a bit odd that combining inner and outer threads doesn't do anything - e.g. 3 outer x 2 inner is pretty much the same as 3x1, and so on). I will look into this a bit more to make sure I didn't do something wrong.
* Looking at run time, in pretty much all cases the inner threading only (blue) line is best (it is much better for memory). Only at 100 templates is 9 outer slightly better over a small range of cores. This suggest we could use inner threading only by default and let the user specify if they want some other combination of inner and outer?

calum-chamberlain:
Thanks for that @chrisdjscott. Do you have a script that you used to profile that? I would like to see if I get the same results for a larger, 80CPU machine.

chrisdjscott:
@calum-chamberlain I don't have a single script to do all the benchmarks. [This](https://gist.github.com/chrisdjscott/4eee5e7b9030a5e23f4487d21cf4c87f) is the script I use to run the test. I set off a number of slurm jobs with different numbers of cores (the script takes two arguments: the number of outer and the number of inner). This works with the latest changes I just pushed.

It prints run times into the standard output and I get the memory usage back from slurm native job profiling - do you use slurm on that machine? Once the run times and memory are in a text file I have a script for plotting the results - can send it if that would help too?

chrisdjscott:
I added the other scripts I've been using on Pan to the gist:

https://gist.github.com/chrisdjscott/4eee5e7b9030a5e23f4487d21cf4c87f

calum-chamberlain:
Thanks for those @chrisdjscott, I haven't had a chance to run them on the machine here, hopefully next week - I will re-write a few things because we are not using SLURM, and so probably won't extract memory usage.

chrisdjscott:
Sounds good @calum-chamberlain - let me know if there is anything I can do to help with this in the meantime.

chrisdjscott:
Trying to work out why no improvement when using both inner and outer threads and it looks like you have to explicitly enable nested parallel loops using `omp_set_nested(1)` (for example see https://www.archer.ac.uk/training/course-material/2014/05/AdvancedOpenMP_Oxford/Slides/L06-NestedParallelism.pdf).

With 100 templates this gives a small improvement. There is an outage on Pan today but I'll rerun the relevant benchmarks when it is back up and update the graphs.

calum-chamberlain:
Ace, thanks for that Chris, I'm still failing to get around to running those tests on our local cluster, but I will try today, with those updates.

calum-chamberlain:
I got around to writing a profiling script in #180, but haven't added cases for different inner and outer threading - maybe tomorrow...? Hopefully.

chrisdjscott:
Results with nested OpenMP loops enabled:

### 100 templates

results_0100_nested

### 200 templates

results_0200_nested

### 1000 templates

results_1000_nested

### 1990 templates 

results_1990

Using nested parallelism gives a small benefit although it still looks to me like defaulting to inner threads only (the blue line) is best (it is only slightly slower for some numbers of threads at 100 templates and memory is always better). The default possibly isn't that important once there is the speed test option from #180.

calum-chamberlain:
I finally started running tests today, I'm going to leave a load running overnight and hopefully have a plot to come back to, it's looking like balancing inner and outer gives the best speed on my 80 CPU machine.

Do you have anything else you want to do with this, or should I give it a review tomorrow and merge?

Thanks again, and sorry for being so slow!

calum-chamberlain:
Of note though, I do appear to be getting occasional (not constant) normalisation errors when using 80 inner threads (with one outer thread). It doesn't seem to happen for any other threading combinations I have tried though. Any thoughts on that? I believe I am running having compiled with mkl, but I will check tomorrow and play with not using mkl to see if it is coming from there...

calum-chamberlain:
So tests failed last night - I'm running into all sorts of issues with the correlations on our cluster, and I haven't figured out why.  The correlations run smoothly on my local machine (and obviously run smoothly on PAN, and on the CI servers), so I'm pegging it as a local issue that I will try and fix with our cluster manager.

Running tests for a range of numbers of templates gives me this:

correlation_speeds_desktop

This was with 15 channel templates at 80 samples long, and 72,000 sample long continuous data. Larger datasets will probably be better with inner threading prioritised.  The legend shows the number of outer threads as the number alongside the method.

More to come when I can figure out what is going wrong with our cluster...

chrisdjscott:
Thanks for taking a look at this. I'm not sure about the normalisation errors - was that with the same dataset as I'm using on Pan? Is it reproducible if you run with same number of cores again?

Strange that it is not running on your cluster. Are you getting some errors or is it just giving wrong results? Is it with MKL or FFTW? If it is a different dataset then might be worth me trying running it on Pan too. I might not get much time to look at this over the next couple of days but can have a closer look after that.

calum-chamberlain:
It looks like it is with both MKL and FFTW, and it isn't with the same data, so it's not really helpful.  I really think it is how I have installed things on our cluster, I imagine I have cocked something up somewhere, so I'm cleaning out and trying again.

On the note of errors, it was giving the expected error for normalisation issues, and I added a print statement to show the numbers for correlation, variance and std, looks like it was in the correlation stage rather than normalisation.  But, I don't think it is worth thinking hard about it, I don't think it is a bug, rather me failing to clean up after myself.

calum-chamberlain:
Happy days, it was indeed my install, unsure quite what, but I suspect installing EQcorrscan as a local user was resulting in some strangeness.  Installing in a virtual env works like a charm.  Graph to follow...

calum-chamberlain:
correlation_speeds_3600_30chan

For an hour of data at 50Hz, and a 30 channel template. Again the number notes the number of outer threads

chrisdjscott:
Cool, did you try with inner threads too or just outer? Did this get rid of the normalisation errors too?

calum-chamberlain:
Ah, sorry, wasn't clear, this all cases use 80 threads total, so if the outer is 1, inner is 80, outer 2, inner 40 and so on.

Normalisation error was definitely my environment, fixed with proper virtual environment.  I'm very happy to see it error so I could work it out rather than silently being weird!

Calum Chamberlain
Sent from mobile.

chrisdjscott:
Cool, makes sense. I will tidy up the code a bit (remove some of the print statements). Is it OK to leave it to default to all inner threads unless the user specifies otherwise, or would you rather some other default? 

For the benchmarks you looked at it seems that a combination of inner and outer is best - I guess it depends on size of data, number of templates and number of channels, so could be complicated to automatically determine the best - I suppose your speed test script could do this though...

calum-chamberlain:
I think leaving the default to be just inner threads is safest, memory usage is much less (which is great, thank you - means we can run many more templates at once which has always been a limit). 

One of the reasons testing took so long was me trying to make the script in #180 call this version (some hacks which will go back into that PR were made to the script).

If you can clean up the code that would be ace, and I have a couple of things that I wanted to tweak I think (adding an ACCEPTED_DIFF define statement rather than using variables in the two functions), which I can either add here, or in a separate PR? Then I think we are good.  Keen to start hammering PAN again in the next couple of weeks - our Kiakoura datasets are coming together so it will be ideal timing. 😄

chrisdjscott:
Awesome, great that it will be useful and good timing. Happy for you to make the changes in here too if that is easiest.

calum-chamberlain:
So I have just added a custom error for errors not due to memory (normalisation errors), and have added a define for `acceptedDiff` so I don't have two separate values for it.
Only comment is, is the [line here](https://github.com/eqcorrscan/EQcorrscan/blob/a0c48ef7803a6f5e2008b5273e0d926b98bbf1c8/eqcorrscan/lib/multi_corr.c#L616) done now, or can we remove the comment?

Also, I think I'm going to take this for a big test run this weekend with a year of data and give it a shake-down for the Kaikoura project :smiley:

chrisdjscott:
That comment can be removed as not relevant any more (I forgot to remove it).

Cool, will be good to see how it goes!

calum-chamberlain:
I ended up having issues running on PAN - I was getting normalisation errors frequently using real data, but when I use the non-internally threaded routines the error did not arise (e.g. the version in develop).

Any chance you could have a look @chrisdjscott - the script I was running is on PAN at:
`/projects/nesi00219/Kaikoura_aftershocks/Kaikoura_afterslip/Kaikoura_detections_hourly_split.py`

It currently loads the compiled version of the #185 branch (compiled having loaded Python/3.6.0-gimkl-2017a). The script takes command line arguments along the lines of:
```bash
python Kaikoura_detections_hourly_split.py --splits n --instance m
```
where n is the number to divide the total dataset duration by, and m is the instance of that group, e.g. when analysing 366 days (as is the case for the script), with n=61, each instance would analyse 6 days of data.

When I ran into issues I had checked out this branch and built as follows:
```bash
cd /projects/nesi00219/Hikurangi_LFES/EQcorrscan
git checkout chrisdjscott-threading-inner
module load Python/3.6.0-gimkl-2017a
python setup.py build_ext --inplace
```
The build-log showed that mkl libraries were being used.

I also noted that PAN is running numpy version 1.11.3, which means that the numpy correlation routine fails because numpy.flip was only introduced in numpy v.1.12.0

chrisdjscott:
Sure, I can have a look at this. So to reproduce the errors I can take the `Kaikoura_detections_hourly_split.py` script and replace the line:

```python
sys.path.insert(0, '/projects/nesi00219/Hikurangi_LFES/EQcorrscan')
```

to point to this branch. Then running (for example):

```
python Kaikoura_detections_hourly_split.py --splits 61 --instance 0
```

Does that sound right? Do you know if the normalisation errors were happening for all instances or only some? Do you still have the output from when you got the errors?

calum-chamberlain:
That sounds right - it looks like I do have some normalisation errors in the #185 branch as well - errors will appear in the logs with the line:
```bash
Error in C code (possible normalisation error)
```

calum-chamberlain:
@chrisdjscott it looks like it may be another issue, probably a data quality issue that I haven't yet caught and should be caught higher up.  Running either the time-domain or fftw routines gives errors, which suggests it isn't an issue in the fftw routine (or, somehow the time-domain and fftw routines have similar errors, which to me suggests a data quality issue).

chrisdjscott:
@calum-chamberlain OK, yes sounds like it could be a data issue if it fails with the time domain method too. I only got as far as running again to see where the normalisation errors were. In case it helps there were some between `2016-11-01T20:59:15.398393Z` and `2016-11-01T21:59:15.373393Z`.

calum-chamberlain:
Thanks for that @chrisdjscott I have managed to find the root of it - gappy data from Durville Island which I padded (which worked okay for old versions), but results in too low variance to be reliable. Now checks and removes traces which have gaps that are > twice the length of the template.

calum-chamberlain:
It looks like a heap of network fails, but I will check - if these tests pass though I'm happy to merge if you are @chrisdjscott ?

chrisdjscott:
Yes I'm happy for it to go in @calum-chamberlain.

PR #172, opened 2017-10-02
Plotting features

cjhopp:
Adds ability to plot detection rate in cumulative_detections (wrapped by Party.plot and Family.plot) as well as specify a date range and a minimum number of detections. Makes the plots more useful when searching through large number of templates over large periods of data.


calum-chamberlain:
Looks like appveyor is playing silly buggers, but otherwise this one should be passing all tests after merge of #175 to sort out circleci's random change to failing.

Do you have any examples of the new plotting/updated docs? Do we need to build the docs for this and check they all work before we merge?

calum-chamberlain:
So we can ignore the codecov patch fail because we haven't written tests for plotting thus far :disappointed: 
I'm building the docs [here](http://eqcorrscan.readthedocs.io/en/cjhopp-plotting-feats/).

calum-chamberlain:
@cjhopp where are we with this - did you check the docs generated and include some plotting examples?

cjhopp:
We areeeeee.......at exactly the same place we were a month ago. Standby for updated docs/examples.

cjhopp:
Alright, @calum-chamberlain . Everything but codecov passes. Lemme know what you think about the examples. Also, do the docs build [here](http://eqcorrscan.readthedocs.io/en/cjhopp-plotting-feats/submodules/autogen/eqcorrscan.utils.plotting.cumulative_detections.html) (for example) automatically? Or is that something I have to trigger on my end?

calum-chamberlain:
Yup, they built automagically, looks like they had no issues on the last build.  Can you put pointers to the docs examples that have changed? Then I will have a check of them and review.
And yup, looks good, shame we still don't test plotting, and something I really want to get going, but not going to happen this year methinks.

calum-chamberlain:
[This](https://pypi.python.org/pypi/pytest-mpl) looks like what we want!

cjhopp:
That does indeed!

Only docs changes should be [here](http://eqcorrscan.readthedocs.io/en/cjhopp-plotting-feats/submodules/core.match_filter.Party.html) and in plotting [here](http://eqcorrscan.readthedocs.io/en/cjhopp-plotting-feats/submodules/autogen/eqcorrscan.utils.plotting.cumulative_detections.html#eqcorrscan.utils.plotting.cumulative_detections)

However, not all of the functions show up in the Party API docs, which was the case before this PR, I think.

calum-chamberlain:
Ah, what are we missing in Party?

calum-chamberlain:
Also, is there anyway to specify bin-size with rate plotting?

cjhopp:
At the moment, it's dynamic based on the size of the date range. Should be simple enough to allow that to be user defined, though.

Quite a few Party functions missing...the ones I changed/added were filter and plot, neither or which appear.

calum-chamberlain:
Can you add them in? Should just have to add them in to `eqcorrscan/doc/submodules/eqcorrscan.core.match_filter.Party.rst` - add both to the `.. autosummary::` list and add an `.. auto method::` statement for each.

calum-chamberlain:
P.S. dynamic is good. 👍

calum-chamberlain:
Looks like somethings gone wrong with this one: http://eqcorrscan.readthedocs.io/en/cjhopp-plotting-feats/submodules/core.match_filter.Party.html#eqcorrscan.core.match_filter.Party.plot

calum-chamberlain:
I think the circleci fail should be fixed in develop now, so updating just to check - its obviously not something you have done.  
Will merge this morning.
Fanks! 👍

PR #173, opened 2017-10-04
Fix memory leak in time-domain

calum-chamberlain:
Memory for the running mean is allocated in the C function for time-domain corr-correlation, but never free-ed, resulting in an annoying memory leak. This PR adds free-ing to both serial and parallel functions.


calum-chamberlain:
This should be updated when #175 is merged, then circleci should pass.  This patch fixes the memory leak on my machines.

PR #177, opened 2017-10-16
Add check for Odroid C2 board

calum-chamberlain:
This PR adds a check for 64-bit ARM architecture.  Previously only 32-bit ARM arch was covered.

Adding this allows compilation of C routines on 64-Bit ARM chips (such as Odroid C2), all it does is omit the `-msse2` compile flag for ARM architectures.


calum-chamberlain:
Strange fail on circle - likely a network thing, running into the same gcc issues as #168

Issue #152, opened 2017-09-07
Determine default correlation function by setup script


calum-chamberlain:
Now that, thanks to @d-chambers, we have a few different correlation functions easily accessible for matched-filter detection, it would be good to help users out by making a script for them that can trial different correlation functions and let them know what the best one for their hardware and dataset is likely to be.

I am thinking about making it write a ~/.eqcorrscan_config file that match_filter then looks for and reads from... not sure how much that would slow it down though, or whether it would be worthwhile or feasible... Open to suggestions.

The way I see it:
- User would call script with size of dataset (number of templates, template length, stream length, number of channels);
- Script would run all available correlation functions for that dataset size, a few times so that numbers can be usefully averaged, and ignoring correlation routines that do not work because hardware is not available (e.g. a GPU routine when we write one);
- Script then tells the user the results and writes the preferred default to a ~/.eqcorrscan_config file (or ~/.eqcorrscanrc, not sure what the nomenclature should be);
- When the user calls match_filter it checks for a ~/.eqcorrscan_config file, if it is there it reads the default from there and, if the xcorr_func and concurrency are not set, uses those in the config file - if there is no config file then it resorts to the general default.

An alternative, longer setup script, could run a range of dataset sizes and write all the results to a config file, then match_filter can query that file and get the fastest routine for the specific dataset size, without user intervention.

Questions @cjhopp @d-chambers @emilyws1 @xiansch @caro-github (and anyone else), is it worth it, are both scripts worth it, or only one? I don't see doing this as something I will do anytime soon, but I think it would be fun and useful eventually.

calum-chamberlain:
This is being worked on in #180

PR #180, opened 2017-10-19
Add correlation speed-test

calum-chamberlain:
This PR is a Work in Progress, and relates to #152. This introduces a script to run all the correlation routines through a given dataset size and plots the time and memory usage for those functions.  It sets a hard limit on the memory usage and will not go in to swap.

An issue that this raises is how to cope with errors in Multiprocessing loops - currently, when an error occurs for one of the workers in a Multiprocessing Pool, the parent process hangs. I tried to get around this by wrapping the function calls in a try/except, but this doesn't appear to have worked.

Open to suggestions!


calum-chamberlain:
Notes from #152:

- [x] User would call script with size of dataset (number of templates, template length, stream length, number of channels);
- [x] Script would run all available correlation functions for that dataset size, a few times so that numbers can be usefully averaged, and ignoring correlation routines that do not work because hardware is not available (e.g. a GPU routine when we write one);
- [x] Script then tells the user the results
- [ ] writes the preferred default to a ~/.eqcorrscan_config file (or ~/.eqcorrscanrc, not sure what the nomenclature should be);
- [ ] When the user calls match_filter it checks for a ~/.eqcorrscan_config file, if it is there it reads the default from there and, if the xcorr_func and concurrency are not set, uses those in the config file - if there is no config file then it resorts to the general default.

An alternative, longer setup script, could run a range of dataset sizes and write all the results to a config file, then match_filter can query that file and get the fastest routine for the specific dataset size, without user intervention.

calum-chamberlain:
This should also test different variations of inner/outer threads for the fftw implementation being worked on in #168

d-chambers:
Hey @calum-chamberlain,

Looking at the work so far it seems like you intend to support running this as a command line utility. I might recommend making this a console script with entry points. [Here](https://stackoverflow.com/questions/774824/explain-python-entry-points) is a good SO question that explains it well, in case you haven't used entry points much. 

Using the "console_scripts" entry point makes it so the user doesn't have to know where this script is located and can simply run something like 
```bash
eqcorrscan-time-correlations
```
maybe get a nice output, and importantly, have the default correlation function set based on the results stored in their home directory.

How cross platform is the memory profile you are doing? I have always used the (psutil)[https://pypi.python.org/pypi/psutil] library for stuff like this to ensure good cross-platform behavior, but if what you are doing is working well no reason to switch. 

Also, there are at least a dozen libraries out there for writing command line utilities. You may not need them for something simple like this but it is good to be aware of them if this ever becomes more extensive, especially if you start using sub-commands. My favorite is [argh](https://pypi.python.org/pypi/argh), but I have also heard good things about [fire](https://github.com/google/python-fire) and [click](http://click.pocoo.org/5/).

calum-chamberlain:
Thank @d-chambers that all looks super helpful - I currently have it set-up so it can be a command-line thing yes, but the command line thing just calls a function with input args, so it could be imported and run.  I don't mind which way we go, but am interested in exploring the console script.  The way EQcorrscan install at the moment (I think) it installs the files in `scripts` onto your path, so your should be able to call them anywhere.  I do think that staying within Python is probably nicer though.

If you are keen for storing the updated default in home directory or similar that is good to know, I think it is a good thing to do, maybe with some info on the dataset size that it was tested for so that match-filter could check and warn when using that default for a very different sized case. Do you think something like a `.eqcorrscan.rc` file?

psutil is definitely cross-platform, I haven't tested memory_profiler on anything other than Ubuntu and OSX yet, but it appears to be stable (I think it uses psutil under the hood), but it does add another dependancy... 

Will also check out those command-line utility things, they sound like they could be useful, thanks!

d-chambers:
If you decide to register this as a console script, or use one of the cli libraries,  you can always still run it from python.  No different than now, just gives the python file an alias of sorts that goes in the users path. I was not aware of the scripts betting added to the user's path,  we wouldn't need console script entry point unless you want it to be known to the user as something different. 

I have seen other python libraries that add config files or other things to the users home directory, as hidden files,  so that is not a bad route to go.  You might also consider putting it in the eqcorrscan module directory,  wherever that lives for the user. They would both work Ok.

eqcorrscan/scripts/correlation_speeds.py
```diff
@@ -0,0 +1,176 @@
...
+
+    :param n_templates:
+    :param n_stations:
+    :param n_channels:
+    :param data_len:
+    :param template_len:
+    :param sampling_rate:
+    :return:
+    """
+    print("Testing correlation functions in %s" % correlate.__file__)
```
d-chambers: Actually you will be testing all registered correlation functions, not just the ones in correlate (although this will probably be true most of the time). Maybe just print the keys of XCOR_FUNCS?

eqcorrscan/scripts/correlation_speeds.py
```diff
@@ -0,0 +1,176 @@
...
+        data_len=data_len, template_len=template_len,
+        sampling_rate=sampling_rate)
+    times = {}
+    memory_use = {}
+    # Limit the memory
+    memory_limit()
+    for corr_func in XCOR_FUNCS.keys():
+        if corr_func == 'default':
+            continue
+        print("=" * 80)
```
d-chambers: Consider creating a small helper function to do this kind of printing, will cut back on repeated code.

eqcorrscan/scripts/correlation_speeds.py
```diff
@@ -0,0 +1,176 @@
...
+from eqcorrscan.utils import correlate
+
+
+def memory_limit():
+    soft, hard = resource.getrlimit(resource.RLIMIT_AS)
+    resource.setrlimit(resource.RLIMIT_AS, (get_memory() * 1024, hard))
+
+
+def get_memory():
+    with open('/proc/meminfo', 'r') as mem:
```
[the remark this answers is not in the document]
calum-chamberlain: Yup, this definitely isn't - psutil will be my friend here, but the function above that uses the `resource` module is more of a pain - `resource` is a standard Python library, but it is Unix only :sob:. I will have to think of another way to limit memory usage on windows.  I may end up writing a wrapper that queries `psutil.virtual_memory` every n seconds, but that will slow things down a bit.  That shouldn't be too much of a issue because that slow-sown should be the same for every function so the times will be systematically shifted, which should not preclude choosing the fastest.

eqcorrscan/scripts/correlation_speeds.py
```diff
@@ -0,0 +1,176 @@
...
+        free_memory = 0
+        for i in mem:
+            sline = i.split()
+            if str(sline[0]) in ('MemFree:', 'Buffers:', 'Cached:'):
+                free_memory += int(sline[1])
+    return free_memory
+
+
+def generate_dataset(n_templates, n_stations, n_channels, data_len,
```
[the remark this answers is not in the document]
calum-chamberlain: Good question, should test other dataset types.

eqcorrscan/scripts/correlation_speeds.py
calum-chamberlain: This was more in there for my debugging, to make sure I was testing the installed functions rather than those on my local path.

eqcorrscan/scripts/correlation_speeds.py
calum-chamberlain: :+1:

calum-chamberlain:
Need to add dependancies if this is what ends up being used:
- [ ] pyyaml
- [ ] memory_profiler

calum-chamberlain:
Now that #168 is merged, this should test for the best arrangement of threads, and should expose that functionality to the user (maybe allow a tuple argument for `cores`?)

calum-chamberlain:
Lots done, but the still to do on this one is:
- [ ] check if randn is appropriate
- [ ] consider checking for optimum function at import, or/as well as within object methods/function
- [ ] write a test/fiish the test for the correlation-speeds script:
  - [ ] Needs both command line and imported tests
- [ ] Add capability to test mixed inner/outer threads for fftw-threaded routine.

PR #182, opened 2017-10-20
Patch low amplitudes

### LINKED ISSUE #181: Data with background amplitudes ~ 10e-7 are not represented well
A user has been playing with active-source data, with very low background amplitudes, and has been getting zero MAD from the correlations - it looks like the FFTW correlation backend does not capture these data accurately - correlations from the time-domain and numpy backends are not affected.

This needs to be fixed reasonable urgently. One way to get around this would be to check the background amplitudes, and if they are below a certain range, multiply them by some power of ten - in practice with these data this works fine.

This should have a test written first with some low-amplitude data to make sure we don't ever re-introduce this behaviour.

calum-chamberlain:
Fixes #181.

This PR changes the `acceptedDiff` value in the fftw correlation backend to a smaller value (the smallest acceptable without incorrectly converting correlations to float) to allow data with lower variance to be used.

It also introduces a warning to explicitly show the user if the variance is below this level.

A test is also added with low amplitudes that previously failed and now does not.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Current
 * Minima not correctly identified in find_peaks in previous versions - bug fixed
 * Compiled peak-finding routine written to speed-up peak-finding.
+* BUG-FIX: changed minimum variance for fftw correlation backend.
 
 ## 0.2.7
 * Patch multi_corr.c to work with more versions of MSVC;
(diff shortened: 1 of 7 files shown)


calum-chamberlain:
To work the network tests I rebased on develop and propose to merge there, then merge develop into master for the patch release.

PR #183, opened 2017-10-20
Change decimation strategy

### LINKED ISSUE #179: Match_filter potting should not always decimate to 10Hz
Currently `eqcorrscan.utils.plotting._match_filter_plot` will try and decimate data to 10Hz without any warning.  This was necessary for large datasets, but is not the case for small datasets.  It becomes an issue for data sampled above 100 Hz because it tried to decimate to 10 Hz in one step (which ObsPy will not allow).

This should be solved by running a check for length and decimating (possibly in a while loop?) until the length is sufficiently short to plot usefully.  Note that this is predominantly a "debug" plot, so I don't care that much about it, but it should not error.

A user running 6kHz data raised this issue.

calum-chamberlain:
This PR fixes #179 

This sets a maximum allowable length in points for the match-filter internal plotting, and runs a while loop to decimate until this is reached.

There is no test for this because plotting is not yet tested.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Current
 * Minima not correctly identified in find_peaks in previous versions - bug fixed
 * Compiled peak-finding routine written to speed-up peak-finding.
+* Change default match-filter plotting to not decimate unless it has to.
 
 ## 0.2.7
 * Patch multi_corr.c to work with more versions of MSVC;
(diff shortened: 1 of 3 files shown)


calum-chamberlain:
Added function for the decimation and a doctest.

calum-chamberlain:
Failures due to missing files during clean-up, always strange failures. Will re-run travis and re-ping Appveyor if they pass.

PR #189, opened 2017-11-09
In core.match_filter.party.rethreshold, do not use list.remove

calum-chamberlain:
This fixes a bug where detections below the threshold were not correctly removed in the `rethreshold` method. Looks like the `remove` method on lists in Python is unreliable, or doesn't do what I expect it to do.

This PR needs a test.


calum-chamberlain:
This now has a test that checks that all new detections are above the threshold.  The threshold is also updated within rethreshold.  Note that these tests would fail for old versions.

PR #192, opened 2017-11-10
Force zeros where there are no data

calum-chamberlain:
Pre-processing pads with zeros before processing - processing can convert these to non-zero, which generates un-real correlations.
This PR tracks zeros added before processing and makes them revert to zeros after processing.

This needs a test.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,10 @@
 * Add filtering detections by date as Party.filter
 * BUG-FIX: Change method for Party.rethreshold: list.remove was not reliable.
 * Add option `full_peaks` to detect methods to map to find_peaks.
+* pre-processing (and match-filter object methods) are now gap-aware and will
+  accept gappy traces and can return gappy traces. By default gaps are filled to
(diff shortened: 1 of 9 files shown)


calum-chamberlain:
This should also incorporate coping with gaps anywhere in a trace, and allow users to pass "gappy" streams to pre_processing functions.
Note that this doesn't seem to affect older correlation functions, but is an issue for the current fftw implementation in `develop`.

calum-chamberlain:
This now has tests that will fail - I think that keeping track of gaps in pre-processing and making sure they stay padded with real zero values (not zero values that have then been processed) should alleviate this.
Another option would be to return masked traces from pre-processing and make match_filter itself cope with them - this would require changes to correlate functions to make them gap-aware. I will only do this if padding with zeros still occasionally raises normalisation errors in fftw correlation...

calum-chamberlain:
Turns out, the errors raised by fftw-concurrent were not real, but were due to incorrect error handling.  This PR is still valid because gaps should be padded with zeros. However, it also fixed the error handling in the C routine.

Issue #186, opened 2017-11-01
Remove Nordic parsing from EQcorrscan


calum-chamberlain:
Now that Obspy 1.1.0 has been released :tada:, which contains Nordic parsing ability (with bugs that were found in EQcorrscan patched in Obspy), we should remove this ability from EQcorrscan.

That will mean removing:
- [x] eqcorrscan.utils.sfile_utils
- [x] eqcorrscan.core.template_gen.from_sfile

It would be good to add a useful error if users try to import such functions.

I will also have to alter the tests, and change dependencies - some tests use Nordic IO, which just needs to be redirected to obspy.read_events.

PR #193, opened 2017-11-15
Remove sfile_utils from EQcorrscan

calum-chamberlain:
This PR removes the `.utils.sfile_util` submodule from EQcorrscan because it is now in obspy (for version 1.1.0), and is being worked on there.  Obspy's version has updates that fix bugs in EQcorrscan's implementation already, and should be where people go to use it.

From the changelog:
* **Removed sfile_utils** This support for Nordic IO has been upgraded and moved
  to obspy for obspy version 1.1.0.  All functions are there and many bugs have
  been fixed. This also means the removal of nordic-specific functions in
  EQcorrscan - the following functions have been removed:
  * template_gen.from_sfile
  * template_gen.from_contbase
  * mag_calc.amp_pick_sfile
  * mag_calc.pick_db
*  All removed functions will error and tell you to use obspy.io.nordic.core.
*  This now means that you can use obspy's `read_events` to read in sfiles.

The reasoning for this removal are as follows:
1. Obspy is the right place for IO modules, EQcorrscan should be concerned with the specifics of earthquake detection and analysis;
2. Maintaining two versions is stupid;
3. Template-generation from nordic (sfiles) remains supported through the `from_meta_file` method - there shouldn't be multiple implementations for the same thing;
4. Fewer odd things in EQcorrscan means easier maintenance and faster testing :tada:


calum-chamberlain:
Closes #186

Issue #153, opened 2017-09-07
Template_gen should be refactored to reduce overlapping code, and simplify function calls


calum-chamberlain:
Almost all functions in `eqcorrscan.core.template_gen` do essentially the same thing:
- Get a catalog (either from a file, from a client, or from a object in memory);
- Get a stream;
- Work out what events are within the stream and trim to useful lengths;
- Process stream chunks;
- Generate template and return a list of templates

Currently there are seven separate functions that do essentially that and have duplicated code within each function.  This should be stream-lined to make it easier to maintain and test these functions. It might be worth implementing something similar to #139?

It is also worth noting for `from_sfile` that this won't be supported soon because sfile reading is now in ObsPy (for version 1.1.0, not yet released), and some fixes have been made upstream to ObsPy that have not been made here.

I'm really looking forward to removing `utils.sfile_utils` at some point too...

yifanyin:
May I also suggest better documentation for the tribe/family/party. And when coming from older versions, how to make use of them with existing template files?

calum-chamberlain:
What would you like to see better documented for tribe/family/party?

Calum Chamberlain
Postdoctoral Fellow, Seismology
Victoria University of Wellington
Sent from mobile.

________________________________

yifanyin:
For example, how to use the new Template object with the template miniseed generated by template_gen(). How to read the detection csv into Family object, etc.

calum-chamberlain:
Thanks @yifanyin, will include details on how to do that - in general you should just be able to populate the fields for the object:
```python
from obspy import read
from eqcorrscan.core.match_filter import Template

st = read("old_template.ms")

template = Template(name='test', st=st, lowcut=2, highcut=10, samp_rate=40, filt_order=4, process_length=3600, prepick=0.05, event=None)
```

You just have to remember the appropriate parameters from when you generated templates. If you have an event it is useful to associate it with the `Template`, but not necessary (until you get to `lag_calc`)

calum-chamberlain:
Alongside this I would like to add a `P_all` and an `S_all` option for the `swin` argument. These should make windows for all channels given starting on the P and S picks respectively.

calum-chamberlain:
Working on this in [this branch](https://github.com/eqcorrscan/EQcorrscan/tree/template-gen-refactor), not ready to make a PR yet - very little refactoring done so far, just adding the options.

PR #199, opened 2018-01-18
Template gen refactor

calum-chamberlain:
Add options to generate multi-channel templates on all P phases or all S phases.

This PR should also refactor template-gen to remove extraneous code as suggested in #153

To do:
- [x] Add `P-all` and `S-all` arguments;
- [x] Refactor and add depreciation warnings;
- [x] Better docs for migration to Tribe/Template objects (see below request)
- [x] Docs for generating Party and Family from csv files.

Request for better docs copied from #153
> For example, how to use the new Template object with the template miniseed generated by template_gen(). How to read the detection csv into Family object, etc.


calum-chamberlain:
Closes #153

PR #200, opened 2018-02-26
Template vs. Detection delay-check & extra stability checks 

flixha:
Hi;
first of all: I really appreciate all of the work that has been put into EQcorrscan, thanks a lot for making such a useful code openly accesible
Also, sorry for this messy pull-request with many small changes, with some requiring more or less discussion. I'm quite a github-newbie, so I'll try my best to have things in smaller chunks next time.

In this PR I have put together several fixes that I needed to run EQcorrscan on data that was partly somewhat messy (spikes, holes, empty traces). That's why I included some extra checks for NaNs after correlation.

Two main things possibly worth discussion:

1.) The despiking-function used a 10 Hz high-pass filter. For my data, the filter made spikes even worse, because it turned "proper" spikes into fourier-transform-artefacts which resemble wavelets. This made the spike-removal (which I only used when the spike filter would else return an error) less effective. These artefacts would still be strong and easily trigger a detection. So I wasn't sure why a high-pass may be useful here?

2.) In lag_calc, when calculating channel-delays, I added some code to check whether the channel that should have been the earliest to detect the event was actually used. This because I had an event which triggered on all channels except this earliest one, even though that earliest channel was present in the continuous data (I suspect some data issues which may have reduced MAD level). This change is also important if one wants to use more stations for lag_calc picking than were used for detection to reduce calculation time. E.g. one may use a template from 10 stations for detection, but extend the template to 50 stations for picking. Then the detection time is corrected to an earlier value according to the earliest channels present during detection and picking.

I'll add more details to the issues/commits in the discussion along the way.

Cheers,
Felix


calum-chamberlain:
Thanks for this @flixha - I will take a look at some of these tomorrow and think about how best to proceed. I looks like you haven't changed that many files, so should be fine in this one PR.
If you can fix the stickler comments (they are all syntax things) that would be really helpful.  It would also be great to have some tests to confirm your changes (although avoid large datasets, maybe for the spike checking use a synthetic dataset?).

flixha:
Thanks! It's getting a bit late now so I'll get to fix these syntax errors (seems to be mostly commenting style) tomorrow. For despiking test I'll check what's easiest, and try to keep it as short as possible.

calum-chamberlain:
Thanks for these @flixha, I think I remember the reason for the high-pass now - when using broadband data, spikes can often be smaller in amplitude than the background long-period noise (as such it would be sensible to use a lower frequency cut).  I am not that fond of a few things in the spike-checking routines though, and wouldn't be at all averse to changing how it works.  However, because we work on copies of the data here, having those horrible, aliased wavelets shouldn't affect the main correlation routines.
Note to self - it might be better to use a [high-order bandstop](http://www.microstrain.com/support/news/how-remove-signal-data-artifacts) than just cut it out.


1.  Main question for you on this is, was the spike checking with the high-pass failing to detect spikes?
2.  Ace, thanks, that will be useful for using more channels in lag-calc than in detection. If you could add a test for that case that would be really helpful. I will try and comment on the changes made in the next few days.

I'm confused by your comment about not having a channel used in detection even though it was in the continuous data - have you been running each channel of a template separately?

calum-chamberlain:
Notes to look into for improving spike detections: EEG literature looks like it might be useful, in particular [this paper on time-frequency domain detection](https://pdfs.semanticscholar.org/9178/010d71804a929160851b522ff453c05fedfb.pdf), and [this on Non-linear Energy Operator](https://www.semanticscholar.org/paper/A-new-interpretation-of-nonlinear-energy-operator-Mukhopadhyay-Ray/a88e489f207158978e97bf6510241b6c90e6822e).

calum-chamberlain:
@flixha It looks like I broke this trying to rebase it - a lot of changes had already been made upstream in develop.  I accidentally updated your develop branch with the current develop, rather than the rebased version, sorry!
I have moved this to a branch within the EQcorrscan repository (flixha-patches) and have made a new pull-request that we can work on: #207

flixha:
Sorry for not getting back to this for a few weeks, I just didn't find the window to look more at this. Yesterday I tried to push (unrelated) changes to my remote, but noticed that these would all appear in this pull request as long as I push to develop, which would be confusing. That brought up the comment on "could not review pull request; no reviewable changes".

So a few comments that I was still owing here:
1. Spikes: Thanks a lot for pointing out the EEG-paper for spike detection. I found astonishingly little literature for spike detection and removal in seismology. I may have confused two issues with the spike-checking and filtering here. I deactivated the high-pass filter when I came across a BH*-channel with 20 Hz sampling rate, which terminated the program because it could only be filtered with a corner frequency up to 9.99... Hz instead of the 10 Hz that were used here. Maybe the only reasonable change right now should be that the sampling rate is taken into account during filtering. I'm not sure any other changes to despiking have gone through enough testing yet.

2. Yes I'll add a test for calculating picks from additional stations/channels than those that were used during match-filter detection as soon as possible.

3. Sorry for the confusing comment, what happened was this: I was checking whether all the earthquakes from which I had created templates were properly detected. I noticed that using lag_calc on one of them returned much fewer picks than it should have (as the template was detecting itself, it should have returned all the picks). The reason turned out to be that due to some data issue (I'm not sure what this was), the closest Z-channel (on which the first pick would reference the detection time) was not part of the detection. However, that Z-channel still existed and lag_calc tried to correlate the template (which did contain that Z-channel) against the window of continuous data defined by the detection time. That data also contained that very Z-channel, so lag_calc referenced the detection time in relation to that Z-channel, even though the detection time was defined from the 2nd earliest pick/channel. 
I wasn't able to track down what caused the correlation to fail on that one channel, but the extra checks that I introduced meant that one may only use one pick instead of a whole event now.

calum-chamberlain:
Hey Felix, 
It was definitely me that broke this - I accidentally overwrote the change you made in your develop branch with the main develop branch, so if you pull from your develop branch you will overwrite your local changes.  I recommend that you force push your local branch (if you still have it) to your github fork:
`git push -f origin develop` Assuming that origin is your github repository.
Otherwise, I have maintained all the changes (I think, apart from some that had to be lost because they conflicted with changes upstream) in the branch in #207. So your changes (including removing the filter) are there and you can play with that.
Sorry again.

1. Me too, I'm happy to try and work on more spike detection stuff at some point, but I'm unsure when I will get the chance - if you do get the chance to work on this, can you create a new branch from the main develop branch and work on it there?
2. Great, can you add that to the branch in #207 please.
3. Thanks for the explanation, I think what you have done there is very useful!

Issue #203, opened 2018-03-20
Correlations calculated during gaps in dev version.


calum-chamberlain:
The current develop branch has a lower threshold for "gaps", which allows correlation at low gains, but also correlates when users have padded with gaps. A comparison of correlation sums for data supplied by a user between the current release and the current develop highlights this issue.
![correlogram_at_gap](https://user-images.githubusercontent.com/10913395/37685516-9eb22002-2cf8-11e8-85e9-b3d7e482c45b.png)
The current master does not calculate correlations for this period, while the current develop branch does.  The user has not used EQcorrscan pre-processing, which means that these gaps are non-zero and therefore missed by the low-gain threshold of EQcorrscan.  Possibly we should be checking for many near-zero values in succession?
![rm33_padding_issue](https://user-images.githubusercontent.com/10913395/37685567-cf59efc8-2cf8-11e8-8863-d2aaa6ea32ce.png)

calum-chamberlain:
This is likely related to #182

calum-chamberlain:
Changing the `ACCEPTED_DIFF` minimum variance for correlation to remove this as an issue only works when `ACCEPTED_DIFF > 1e-10` - currently set to `1e-15` in dev.  Going to this level will mean that active source data (as in #182) would not be processed.
Options as I see it:
1. Change to higher `ACCEPTED_DIFF` (which will tell the user if correlations are not computed and suggest they increase the gain;
2. Keep everything as is and add docs, and possibly an additional warning when the variance is low, but non-zero;
3. Add in checks for successive similar values - not sure how this would work, and it would likely add in a lot of extra overhead.

In short, I think this isn't a "real' data issue, in that it is trying to get around a data issue without understanding the consequences (which is likely my fault, this should be better [documented](http://eqcorrscan.readthedocs.io/en/latest/submodules/autogen/eqcorrscan.core.match_filter.match_filter.html#eqcorrscan.core.match_filter.match_filter)) - so I think option 2 is probably best (and certainly easiest to manage).

calum-chamberlain:
Closed by #204

PR #204, opened 2018-03-28
Warn when variance is low, and add docs

calum-chamberlain:
This PR implements option 2 from #203.  Users are warned when the variance of the data is worryingly low (not non-zero) because this might be due to gaps padded after filtering. This copes with a change in behaviour from 0.2.7.


calum-chamberlain:
To do:
- [ ] Add note in pre-processing about recommended gap filling (e.g. do it after filtering if you are going to do it yourself!).

PR #207, opened 2018-04-05
Template vs. Detection delay-check & extra stability checks

calum-chamberlain:
Accidentally overwrite @flixha develop branch, so moving this PR here.  Copied info from #200:

> Hi;
> 
> first of all: I really appreciate all of the work that has been put into EQcorrscan, thanks a lot for making such a useful code openly accesible
> Also, sorry for this messy pull-request with many small changes, with some requiring more or less discussion. I'm quite a github-newbie, so I'll try my best to have things in smaller chunks next time.
> 
> In this PR I have put together several fixes that I needed to run EQcorrscan on data that was partly somewhat messy (spikes, holes, empty traces). That's why I included some extra checks for NaNs after correlation.
> 
> Two main things possibly worth discussion:
> 
> 1.) The despiking-function used a 10 Hz high-pass filter. For my data, the filter made spikes even worse, because it turned "proper" spikes into fourier-transform-artefacts which resemble wavelets. This made the spike-removal (which I only used when the spike filter would else return an error) less effective. These artefacts would still be strong and easily trigger a detection. So I wasn't sure why a high-pass may be useful here?
> 
> 2.) In lag_calc, when calculating channel-delays, I added some code to check whether the channel that should have been the earliest to detect the event was actually used. This because I had an event which triggered on all channels except this earliest one, even though that earliest channel was present in the continuous data (I suspect some data issues which may have reduced MAD level). This change is also important if one wants to use more stations for lag_calc picking than were used for detection to reduce calculation time. E.g. one may use a template from 10 stations for detection, but extend the template to 50 stations for picking. Then the detection time is corrected to an earlier value according to the earliest channels present during detection and picking.
> 
> I'll add more details to the issues/commits in the discussion along the way.
> 
> Cheers,
> Felix


eqcorrscan/core/lag_calc.py
```diff
@@ -227,17 +236,18 @@ def _channel_loop(detection, template, min_cc, detection_id, interpolate, i,
...
     event.comments.append(Comment(text=ccc_str))
     if used_chans == detect_chans:
         if pre_lag_ccsum is not None and\
-           checksum - pre_lag_ccsum < -(0.30 * pre_lag_ccsum):
+           checksum - pre_lag_ccsum < -(0.3 * pre_lag_ccsum):
             msg = ('lag-calc has decreased cccsum from %f to %f - '
                    % (pre_lag_ccsum, checksum))
-            # warnings.warn(msg)
-            raise LagCalcError(msg)
+            warnings.warn(msg)
```
calum-chamberlain: Changing this results in failed tests - lag-calc should never reduce the correlation sum if the same channels are used. What was the reason for changing this?

eqcorrscan/core/match_filter.py
```diff
@@ -4155,6 +4156,12 @@ def match_filter(template_names, template_list, st, threshold,
                       stream[0].stats.starttime.datetime.strftime('%Y%j')]),
             4, debug)
         if all_peaks[i]:
+            if len(all_peaks[i]) > 1000:
```
calum-chamberlain: This seems like quite an arbitrary threshold (not that there aren't other arbitrary thresholds in EQcorrscan), but it seems like something that should be kept outside of EQcorrscan and in a users own script. I have cases where I have thousands of detections in a single-stream and would rather not have this warning.

eqcorrscan/utils/despike.py
```diff
@@ -58,8 +58,8 @@ def median_filter(tr, multiplier=10, windowlength=0.5,
     # Note - might be worth finding spikes in filtered data
     filt = tr.copy()
     filt.detrend('linear')
-    filt.filter('bandpass', freqmin=10.0,
-                freqmax=(tr.stats.sampling_rate / 2) - 1)
+    # filt.filter('bandpass', freqmin=9.999,
```
calum-chamberlain: It would be good to have a separate conversation and pull request to improve despiking, so can we leave this out for now?

calum-chamberlain:
Hey @flixha I made some changes here as noted in the commits, I think everything passes now and it would be great to merge this.  Are you happy with what I have done?
@cjhopp a quick review would be good too.

PR #225, opened 2018-06-05
WIP: Add logging

calum-chamberlain:
Work in progress on converting print statements in match-filter functions to logging. This should allow the user to provide their own logger and customise how much information is output, how it is given, and where it goes (e.g. to a file, to screen, or both).

This is at the request of @nackerley, although I have meant to do this for ages.

At the moment this does not work (as will be evidenced by failing tests).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 * Cleaned imports in utils modules
 * Removed parallel checking loop in archive_read.
 * Add better checks for timing in lag-calc functions (#207)
+* Convert most print statements in match-filter functions to logging calls. The user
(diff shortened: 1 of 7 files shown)


calum-chamberlain:
So this doesn't work because it's not simple to propagate loggers through multiprocessing. It looks like I may [not be able to do this in a cross-platform way](https://stackoverflow.com/questions/641420/how-should-i-log-while-using-multiprocessing-in-python#comment12070361_894284).

I'm going to try a different way [here](https://github.com/eqcorrscan/EQcorrscan/tree/logging_attempt_2).

PR #227, opened 2018-06-05
Remove gap threshold

### LINKED ISSUE #224: Gap threshold in `Tribe.client_detect` seems very short
`match_filter.Tribe.client_detect` throws out traces with gaps longer than twice the length of the template: https://github.com/eqcorrscan/EQcorrscan/blob/master/eqcorrscan/core/match_filter.py#L2690-L2700

This seems like a very short gap, and gaps should be handled by `eqcorrscan.utils.process`, although @nackerley has reported some issues here when removing the gap thresholding - many `low variance` warnings are thrown by the underlying C code during correlations, which is often symptomatic of gaps padded with zeros before processing.

I should experiment with removing this thresholding, and allowing a user defined gap threshold.

calum-chamberlain:
Closes #224 

The gap threshold seemed a little arbitrary and very short. @nackerley pointed this out when working on his own data and making his own method that allowed him to take data from multiple clients. Tests pass locally, but I haven't done anything special.

This needs:
~~- [x] A test with known gappy data to ensure that we don't get warnings of low variance from the underlying C-code~~
~~- [ ] Make a test that actually captures that output, or pass it back to Python properly...~~

These are handled by the `correlate.py` function now (see #230)

- [x] Test for gappy data from client - I don't know where this might come from.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 * Cleaned imports in utils modules
 * Removed parallel checking loop in archive_read.
 * Add better checks for timing in lag-calc functions (#207)
+* Removed gap-threshold of twice the template length in `Tribe.client_detect`, see
+  issue #224.
 
 ## 0.3.0
 * Compiled peak-finding routine written to speed-up peak-finding.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Current
 * Cleaned imports in utils modules
 * Removed parallel checking loop in archive_read.
+* Add better checks for timing in lag-calc functions (#207)
 
 ## 0.3.0
 * Compiled peak-finding routine written to speed-up peak-finding.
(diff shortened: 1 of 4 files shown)


calum-chamberlain:
Some of this work is being undertaken in #230 (in the underlying C code and passing warnings back to Python).  This should not be merged until after #230.

calum-chamberlain:
@nackerley, do you have a client (and time-period) that I can point that has gappy data so that I can write the test for this and merge it? The work in #230 should, I think fix most of the issues you were having with gaps, and I added more thorough tests for gaps in the correlations, and warnings are now passed back to Python rather than being raised from the C library.  This should make #229 (using logging) more useful as well.

calum-chamberlain:
Right, in grockeling around I found some gappy data and wrote some tests, they pass locally, if they pass on ci I will merge and release 0.3.1 🎉

calum-chamberlain:
Mac OS fails on travis are unrelated (travis issues and network issues).

PR #229, opened 2018-06-08
WIP: Use logging

calum-chamberlain:
This superseeds #225.

This attempt does not pass `Logger`'s between processes, rather using the root logger.  The user should be able to set up the root logger themselves, but it needs to be *documented*.

Currently this probably overloads the settings of `level` and `format` that a user would set internally, open to discussion about whether this should be done, or, if there is a way to check if the level and format has already been set.  If it has already been set, then it doesn't make sense to reset it.

Setting up the root logger can be done something like:
```python
import logging

rootLogger = logging.getLogger()
rootLogger.setLevel("debug") # Options are debug, info, warning, error, critical
# Create a fileHandler to log output to file
fh = logging.FileHandler('eqcorrscan.log')
fh.setLevel(logging.DEBUG)
# create console handler with a higher log level
ch = logging.StreamHandler()
ch.setLevel(logging.ERROR)
# create formatter and add it to the handlers
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
fh.setFormatter(formatter)
ch.setFormatter(formatter)
# add the handlers to the logger
rootLogger.addHandler(fh)
rootLogger.addHandler(ch)
```

So, things to do on this are:
- [ ] Check whether the logger has already been set-up and don't overload those settings internally
- [ ] Add docs on how to set up a logger - link to [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html)
- [ ] Convert remaining print statements to logging statements in the `core` functions and those called by the `core` functions
- [x] Consider how to better warn of low variance from C-code (see #224)

This PR also adds the option of multiple `swin` arguments and removes the error when data are too short (instead data are removed). This needs to be:
- [ ] documented;
- [ ] tested.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,11 @@
   is the same value repeated (#224, #230).
 * BUG-FIX: template generation `from_client` methods for swin=P_all or S_all
   now download all channels and return them (as they should). See #235 and #206
+* Change standard debug printing to logging (#229).
+* Change from raising an error if data from a station are not long enough, to
+  logging a critical warning and not using the station.
+* Add ability to give multiple `swin` options as a list. Remains backwards
+  compatible with single `swin` arguments.
 
 ## 0.3.0
 * Compiled peak-finding routine written to speed-up peak-finding.
--- a/eqcorrscan/core/template_gen.py
+++ b/eqcorrscan/core/template_gen.py
@@ -41,7 +41,6 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import warnings
 import numpy as np
 import copy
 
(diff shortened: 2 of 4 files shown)


calum-chamberlain:
Note to self: [This section of the logging How-To](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) seems like it will be useful.

calum-chamberlain:
Or [this](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) might be more useful. At the moment the logger name is fixed as the rootLogger, which is not so nice.

calum-chamberlain:
Closing this because this isn't how I want to do it.

PR #231, opened 2018-06-12
Start adding FMF docs

### LINKED ISSUE #217: Add how-to wrap Fast-Matched-Filter for EQcorrscan
In cases where memory is tight, or where threads are bountiful (including NVIDIA GPU threads), [FastMatchedFilter](https://github.com/beridel/fast_matched_filter) has very efficient time-domain correlation routines. In these cases the time-domain routines are more efficient than our frequency domain routines.

It is simple to wrap the FastMatchedFilter correlation routines so that they can be called from EQcorrscan, and the docs should add a how-to.

At the moment FastMatchedFilter is not distributed by any other mechanism than github, but it is still simple to install (with a Python setup.py file, and modular structure).  The guide should therefore include a quick note on how to install, and a snippet for wrapping the appropriate functions.

calum-chamberlain:

### What does this PR do?

This PR adds docs to demonstrate using [Fast Matched Filter](https://github.com/beridel/fast_matched_filter) code from EQcorrscan.

This PR closes #217 

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `develop` for bug fixes.
- [x] All tests still pass.
- [x] Significant changes have been added to `CHANGELOG.txt`.
- [x] First time contributors have added your name to `CONTRIBUTORS.txt`.
- [x] Docs finalised and checked on readthedocs. Pages to be checked:
  - [x] [utils.correlate](https://eqcorrscan.readthedocs.io/en/fmf-docs/submodules/utils.correlate.html)
  - [x] [fast-matched-filter](https://eqcorrscan.readthedocs.io/en/fmf-docs/fast-match-filter.html)
  - [x] Consider how best to link to the fast-matched-filter page and where it should be in the toctree.


calum-chamberlain:
The docs are up now, but in testing locally I get zeroed correlations so I assume I am doing something wrong...

Note that this is unlikely to be tested as part of the doctests, because it would require installation of FMF on the CI which adds an extra level of complexity.

calum-chamberlain:
To finish this:
- [ ] Get the green-light from @beridel
- [ ] Work out why the code doesn't generate correlations (probably something about shape).

beridel:
I haven't yet had the time to try it out myself, but that looks good to me! My only comment would be: it might be worth it to include the wrapper function that you cite in the readme. You could include some test when called to see if FMF is installed, and if not, it would just spit out an error.

calum-chamberlain:
Thanks for that @beridel - Good point on the wrapper, and that would be simple to implement, however I'm kinda keen not to include the wrapper in EQcorrscan for three main reasons:
1. I don't want to write the extra tests required (although this really wouldn't be that much of a pain and would lead to better code in the long-run);
2. Because we don't use weighting in EQcorrscan (yet), but I want the user to be able to make use of that weighting: the wrapper as it is allows the users to define some dictionary of standard weighting for some channels, which might be a reasonable compromise?;
3. I do not want to take undue credit for fast-matched filters great work, I fear that if I include a wrapper then people might forget the effort that you guys put in - I could however make the wrapper print out a note saying something like: `You are using FMF, please remember to cite: ...`?

What do you think about those (1 is my problem and I can deal with it)?

PR #245, opened 2018-07-04
Greater parallel control in match-filter

calum-chamberlain:

### What does this PR do?

This PR:

1. adds a `process_cores` argument to match_filter functions to limit the processes spooled up by `pre_processing` and `findpeaks`, which can be quite memory intensive (as currently written), whereas the correlation functions are more optimised now and can run at higher process-counts (still controlled by `cores` argument).  
2. allows `cores_outer` to be passed through to the fftw correlation routines, allowing more control there (as needed by #180).
3. allow `Party` objects to be written without a detection catalog (the default is still to write the catalog) as this can be quite time consuming and memory intensive. If catalogs are not present in the `Party` archive upon reading it, the catalog will be recomputed.  For large catalogs this may be faster than reading the catalog from disk.
4. fix a **BUG** whereby when templates with duplicate channels were run alongside templates without duplicate channels, those without duplicate channels had duplicate picks in their events with spurious pick-times.

No relevant issues, but this is required by #180 

### PR Checklist
- [x] All tests still pass.


calum-chamberlain:
Needs test on writing and reading from nordic and sc3ml formats.

calum-chamberlain:
Appveyor fails are unrelated (but changing how the tests work so that more are run everywhere highlighted this) - tracked in #247 Other fails are unrelated, merging.

Issue #234, opened 2018-06-16
Correlation routines could return peaks


calum-chamberlain:
**Is your feature request related to a problem? Please describe.**
Find peaks now uses a C routine for finding peaks as it is, we may as well call that and just return the peaks to python rather than the full correlation sum vector. This would allow openmp parallelism rather than expensive python multiprocessing.

**Describe the solution you'd like**
Add boolean variable `find_peaks` to correlation functions and make this the default for calling from within match filtering

calum-chamberlain:
I'm going to close this - I could do this, but I would need to do sorting in C (`qsort` would be the key here). #249 did a good job of speeding up finding peaks, and getting the full correlation function back to Python is useful to debugging.

PR #249, opened 2018-07-06
Find peaks speedup

calum-chamberlain:

### What does this PR do?

Uses a compiled peak-finding routine, which can eventually be ported to the correlate functions so that peaks can be returned rather than the full correlation array (not done here for compatability reasons)

#234 

### PR Checklist
- [x] All tests still pass.

### To Do:
- [x] Use the openmp loop rather than multiprocessing
- [x] Do the sort and decluster in C?


calum-chamberlain:
In testing this speeds up the peak-finding by ~2 times. It is also more memory efficient because it doesn't copy the correlation sum array.

calum-chamberlain:
Bumping to 0.4.0 because this is more of a feature than bug-fix.

calum-chamberlain:
I have added parallel declustering (using openMP) and numpy sorting (which is *much* more efficient than my previous zip/python sort combo) and now the compiled versions are around 50 times faster than the old Python versions. :tada: 

In testing this works well on both Windows and Ubuntu - travis seems to be having issues getting the toolchains it needs from conda for some reason.

calum-chamberlain:
Looks like the toolchain things are probably related to [this](https://twitter.com/condaforge/status/1083392304042463233) - change of compilers, so hopefully this will be completed soon.

calum-chamberlain:
MSVC 8.0 (2008) does not define `llabs`.  I can't be bothered to waste time making this work for >10 years out of date compilers, so I'm just going to stop testing using it. If anyone runs into an error when linking:
```
... error LNK2019: unresolved external symbol llabs referenced in ...
```
then this is likely why and you should update your MSVC.

PR #253, opened 2018-07-13
Do not force day cuts

calum-chamberlain:

### What does this PR do?

Changes the behaviour of `template_gen` (and therefore `Tribe.construct`) for data that are day-long length, but do not start at the start of a day.  Previously data that started at say 5pm, were padded to start at midnight (17 hours before), and trimmed at the end of the day.  I'm fairly sure this isn't what anyone would want.

Now the data are checked to work out if the data actually do start near a day-break and then uses dayproc, otherwise it uses shortproc, which is a touch slower, but doesn't enforce as much.

### PR Checklist
- [ ] All tests still pass.
  - [ ] Test for day-long data that do not start at day-break - test should check that templates that are in the next day are made (they weren't in the old version because data were trimmed at the end of day).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@
 * Add `select` method to `Party` and `Tribe` to allow selection of a 
   specific family/template.
 * Add ability to "retry" downloading in `Tribe.client_detect`.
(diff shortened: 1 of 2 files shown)


calum-chamberlain:
Just network fails, merging.

PR #260, opened 2018-08-09
Patch process-len gaps for client_detect

calum-chamberlain:
### What does this PR do?

* Enforce pre-processing even when no filters or resampling is to be done to ensure gaps are properly processed (when called from `Tribe.detect`, `Template.detect` or `Tribe.client_detect`)
* BUG-FIX in `Tribe.client_detect` where data were processed from data  one sample too long resulting in minor differences in data processing (due to difference in FFT length) and therefore minor differences   in resulting correlations (~0.07 per channel).

Tests were sporadically failing when different lengths of data were download on the CI systems.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,12 @@ install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       export OS="MacOSX";
       export py=$PYTHON_VERSION;
-      brew update;
+#      brew update;
       brew install gcc6;
       export CC=gcc-6;
       export CXX=g++-6;
-      brew install fftw;
-      brew list;
+#      brew install fftw;
+#      brew list;
     else
       export OS="Linux";
       export py=$PYTHON_VERSION;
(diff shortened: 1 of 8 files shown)


calum-chamberlain:
NCEDC is down.

calum-chamberlain:
Now the FFTW ftp is down, so appveyor builds fail.

calum-chamberlain:
It looks like the FFTW routines give different enough results with different numbers of threads to result in slightly different (~0.01) normalised correlations values. This is only cropping up on two of the travis instances (Linux py 2.7, no mkl, and OSX py 3.5 no mkl). So far I haven't been able to replicate this, but it seems consistent on those instances.

calum-chamberlain:
I am guessing that the differences stem from differences in how FFTW makes plans for shared-memory parallel versus single-threaded.  More tests need to be done! 

The differences in correlations are pretty minor, but if there is no way around them, then those differences should be noted and made obvious to users, e.g. with a note in the docs.  The note should have a description of where the differences arise.

If this is the real source of the differences, it might also be possible to define the number of threads that FFTW uses in the header file so a user can force single-threaded transforms if they decide that this is an issue for their use-case, or maybe some global vairable (`FFTW_MAXTHREADS` that if isn't set just gets set to `OMP_N_THREADS`)?

calum-chamberlain:
Still a bug on OSX that I'm trying to track down. Haven't managed to replicate it anywhere.

PR #267, opened 2018-08-22
ci specific setup

calum-chamberlain:
Ci-specific set up for test-script.


calum-chamberlain:
This probably shouldn't use a symlink and just specify where eqcorrscan is and test that - Windows doesn't like symlinks, but there are other issues in windows as well (with using pytest multiple times).

PR #268, opened 2018-08-26
Don't always use test-script

calum-chamberlain:
Use py.test and report coverage properly, but also test the test-script.

Once merged, develop should be merged into master.


calum-chamberlain:
criceci is failing on subspace tests, for which nothing has changed since the last passing test - it looks like the data have changed that are downloaded from GeoNet.  If this is the same tomorrow I will update the test-set so that this doesn't fail.

calum-chamberlain:
No clear reason for fail - master also fails now, but did not fail before. There have been no changes to the codebase that should have caused this change - the plots look good, detections are similar, so I have just updated the comparison files.

PR #266, opened 2018-08-21
Speed-up distance clustering

calum-chamberlain:
### What does this PR do?
Uses a c-extension to calculate the distance matrix for distance clustering.  Improves speed (using openmp looping), and makes the Python interface more standardised between different clustering methods.

### Why was it initiated?  Any relevant Issues?

Distance clustering is unnecessarily slow.

### PR Checklist
- [x] `develop` base branch selected?
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGES.md`.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.~~

PR #285, opened 2018-12-24
Implement SEC-C style internal chunking of frequency domain correlations

calum-chamberlain:
### What does this PR do?

This PR is inspired by the [SEC-C paper](https://pubs.geoscienceworld.org/ssa/srl/article/566118/super-efficient-cross-correlation-sec-c-a-fast). It implements internal chunking of the FFTs of the data. Initial testing shows this to be faster than running really large FFTs.  It should also be much more memory efficient because the size of FFTs can be reduced. At the moment this is controlled with an `fft_len` kwarg on the fftw correlation functions.  This can be passed through from the matched-filter functions.

[Senobari et al. (2018)](https://pubs.geoscienceworld.org/ssa/srl/article/566118/super-efficient-cross-correlation-sec-c-a-fast) make the point well that EQcorrscan can be very costly in memory.  Currently the way around this is to use either shorter processing lengths, or group templates.  Neither of these options is very efficient. They present a simple alternative whereby many shorter FFTs can be computed for the correlations.  This does not effect accuracy (as using an incorrect/different processing length between template and data would), and allows the template FFTs to be cached while looping through chunks of continuous data. A side effect of this is that longer streams of data could be worked on efficiently.

### To do:
- [x] Document this behaviour;
~~- [ ] Estimate the most efficient `fft_len` on the fly given the memory restrictions of the system;~~
~~- [ ] Further parallelism could be enabled, e.g. the current `outer_core` parallelism could be changed to work on the loop over chunks of continuous data;~~
- [x] Testing using #180 would be good, some graphs demonstrating the different memory and time requirements would be nice.
- [x] Wait until #266 is merged, which makes changes to the C functions which will require some tweaking to merge with this.

- Long-term, this could be memory efficient enough to be ported to the GPU, which could allow for some serious desktop speed-ups.

### PR Checklist
- [x] All tests still pass.


calum-chamberlain:
There is an indexing error now with this that needs to be resolved.  This is readily apparent in the `correlate_test.py::TestArrayCorrelateFunctions::test_single_channel_similar` test.

calum-chamberlain:
So it looks like chunking is good, faster, more efficient at loading CPUs, and less memory intensive.  It also passes tests and gives the same results.  The [gist here](https://gist.github.com/calum-chamberlain/597e91066abcecbb830c91773112bed4) shows some of the profiling I ran. I ran a range of other dataset sizes and found that an fft-length of 2**13 was always fastest on my machine. I am tempted to auto-set the fft-len to this.

PR #324, opened 2019-08-28
Develop add plotdir to lag_calc function

imankahbasi:
### What does this PR do?
       for save plot in lag_calc function
### PR Checklist
- [ ] All tests still pass.
- [x] First time contributors have added your name to `CONTRIBUTORS.md`.

### DIFF
--- a/eqcorrscan/core/lag_calc.py
+++ b/eqcorrscan/core/lag_calc.py
@@ -12,15 +12,17 @@ import numpy as np
 import scipy
 import logging
 
-from multiprocessing import Pool, cpu_count
 from collections import Counter
 
-from obspy import Stream
+from obspy import Stream, Trace
 from obspy.core.event import Catalog
 from obspy.core.event import Event, Pick, WaveformStreamID
 from obspy.core.event import ResourceIdentifier, Comment
 
-from eqcorrscan.utils.plotting import plot_repicked, detection_multiplot
+from eqcorrscan.utils.correlate import get_stream_xcorr
+from eqcorrscan.core.match_filter.family import Family
+from eqcorrscan.core.match_filter.template import Template
+from eqcorrscan.utils.plotting import plot_repicked
 
 
 Logger = logging.getLogger(__name__)
@@ -69,9 +71,10 @@ def _xcorr_interp(ccc, dt):
         last_sample += 1
     num_samples = last_sample - first_sample + 1
     if num_samples < 3:
-        raise IndexError(
+        Logger.error(
             "Fewer than 3 samples selected for fit to cross correlation: "
             "{0}".format(num_samples))
+        return np.argmax(cc) * dt, np.amax(cc)
     if num_samples < 5:
         Logger.warning(
             "Fewer than 5 samples selected for fit to cross correlation: "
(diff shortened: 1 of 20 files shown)


eqcorrscan/core/lag_calc.py
```diff
@@ -481,6 +483,10 @@ def lag_calc(detections, detect_data, template_names, templates,
         To generate a plot for every detection or not, defaults to False
     :type parallel: bool
     :param parallel: Turn parallel processing on or off.
+    :type plotdir: str
+    :param plotdir:
+        Path to plotting folder, plots will be output here in a folder of date
```
calum-chamberlain: It looks like the doc-strong doesn't match what directory is being created. Can you change this to match the code?

eqcorrscan/core/lag_calc.py
```diff
@@ -631,6 +637,11 @@ def lag_calc(detections, detect_data, template_names, templates,
                             template_plot.remove(tr)
                     plot_repicked(template=template_plot, picks=event.picks,
```
calum-chamberlain: This takes the usual plotting arguments - currently the default is to show the figure.  I think we should keep the default to be as is (showing the figure), however, if plotdir is given, then we should pass a dictionary of kwargs. Something like:
```python
if plotdir:
    if not os.path.isdir(plotdir):
        os.makedirs(plotdir)
    savefile = "{plotdir}/{rid}.png".format(plotdir=plotdir, rid=event.resource_id.id)
    plot_kwags = dict(show=False, save=True, savefile=savefile)
else:
    plot_kwargs = dict(show=True)
plot_repicked(template=template_plot, picks=event.picks, det_stream=plot_stream, **plot_kwargs)
```
This would also mean that matplotlib doesn't have to be imported here, and would make use of the plotting api a bit better.

CONTRIBUTORS.md
```diff
@@ -5,3 +5,4 @@
 * Shanna Chu
 * Derrick Chambers
 * Chris Scott
+* iman kahbasi
```
calum-chamberlain: Can you capitalise your name please to match the others?

eqcorrscan/core/lag_calc.py
```diff
@@ -431,7 +433,7 @@ def _prepare_data(detect_data, detections, template, delays,
 def lag_calc(detections, detect_data, template_names, templates,
              shift_len=0.2, min_cc=0.4, horizontal_chans=['E', 'N', '1', '2'],
              vertical_chans=['Z'], cores=1, interpolate=False,
-             plot=False, parallel=True):
+             plot=False, plotdir='.', parallel=True):
```
calum-chamberlain: I think the default should be `None` and if plotdir is None then the figure should be shown, which is currently the expected behaviour. See the comment below about how plotdir could be used.

calum-chamberlain:
It looks like your develop branch is not fully up-to-date with the current develop branch.  Can you pull the latest changes from develop into your branch?  The plotting is now handled within `xcorr_pick_family`, and the `plotdir` argument should be passed down to that function.

eqcorrscan/core/lag_calc.py
imankahbasi: Yes. I prefer this method too. But i want to make it look like match_filter.

eqcorrscan/core/lag_calc.py
imankahbasi: Hi, dear chamberlain.
Thank you for your guide.
I learn something more from your comment and i thanks very much.
I will try again in correct way.
I hope this time everything will be correct.

PR #329, opened 2019-09-03
Add path to save plot and some improvement in draw a figure

imankahbasi:
### What does this PR do?
add feature to save figure of plot in a directory
improve shape of figure
### PR Checklist
- [x] `develop` base branch selected?
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGES.md`.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.~~

PR #330, opened 2019-09-04
Add plotdir throughout

calum-chamberlain:
### What does this PR do?

Adds a standardised `plotdir` argument to:
- `core.template_gen.template_gen`,
- `core.matched_filter.match_filter.match_filter`,
- `core.matched_filter.family.lag_calc`
- `core.matched_filter.party.lag_calc`
- `core.matched_filter.template.construct`
- `core.matched_filter.template.detect`
- `core.matched_filter.tribe.detect`
- `core.matched_filter.tribe.client_detect`
- `core.matched_filter.tribe.construct`

This also adds some of the minor changes in #329.

#329 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,8 @@
 * Refactor of catalog_to_dd functions:
   - Speed-ups, using new correlation functions and better resource management
   - Removed enforcement of seisan, arguments are now standard obspy objects.
-* Add plotdir to lag_calc.xcorr_pick_family
-  - Use for save plot. Figures will save with name of events' id
+* Add plotdir to lag-calc, template construction and matched-filter detection
+  methods and functions.
 * Wholesale re-write of lag-calc function and methods. External interface is
   similar, but some arguments have been depreciated as they were unnecesary.
   - This was done to make use of the new internal correlation functions which
--- a/eqcorrscan/core/lag_calc.py
+++ b/eqcorrscan/core/lag_calc.py
@@ -292,7 +292,8 @@ def xcorr_pick_family(family, stream, shift_len=0.2, min_cc=0.4,
                checksum - detection.detect_val < -(0.3 * detection.detect_val):
                 msg = ('lag-calc has decreased cccsum from %f to %f - '
                        % (detection.detect_val, checksum))
-                raise LagCalcError(msg)
+                Logger.error(msg)
+                continue
         else:
             Logger.warning(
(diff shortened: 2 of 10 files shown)


imankahbasi:
it looks very good.
isn't it better write os.makedirs in _finalise_figure() for one time and remove it from all other functions?
And instead of make & update "plot_kwargs" update main "kwargs"?
in this way user can define kwargs and update inside code, both of user kwargs and plot_kwargs will be use, and kwargs can pass from function to another inside function.

calum-chamberlain:
So the way you have it seems confusing to me:
1. If you pass a `savefile` to either of these two plots it gets used as a directory name and the actual filename is changed. This is different to every other plotting function there, so seems like a bad idea
2. In this way the user has no control over the actual name of the plot.
3. If a user passes `savefile` in the normal way (like "directory/filename.format") then a directory with a .format is created. This is strange
4. You are passing the `plotdir` kwarg to the plotting functions, but this isn't actually used by the plotting functions as far as I can tell.

The way I would like it to work is:
1. Plotting API is left relatively as-is, plots should have a similar interface to one another (which I am generally bad at).  The plotting API should be seen as something to be used by a user in scripts;
2. Any additional things required by other functions in EQcorrscan should handle those additional things, whether this be by making directories, or post/pre processing figures.

imankahbasi:
Hi, dear Chamberlain.
answer to 3)
with this method in _finalise_figure()
    if save:
        if '/' in savefile:
            plotdir = '/'.join(savefile.split('/')[:-1])
            if not os.path.isdir(plotdir):
                os.makedirs(plotdir)
savefile split with '/' and result for  "directory/filename.format" is not ".format". it is only "directory".

but let's move on.
Please tell me how can i edit #329  like things you would like?

calum-chamberlain:
While that is true for all other functions, this section:
```python
    plotname = '{}_SNR.png'.format(signal[0].stats.starttime)
    if "savefile" in kwargs.keys():
        savefile = "{}/{}".format(kwargs.get("savefile"), plotname)
```
takes a users `savefile` and adds an additional filename on the end, so the `savefile` passed to `_finalise_figure` is not the same as passed in.

RE what to do about #329: As I said, if this PR gets the result you were after (being able to pass a `plotdir` argument to key functions and making some small changes to figure format) then I would close #329 and I will merge this PR later today.

PR #334, opened 2019-09-21
New function for visual check in families between detections and parents.

imankahbasi:
### What does this PR do?
Function to plot a graph to show all detections in one day, and comparison between them.
Plot detections for each family with separate colors.
### PR Checklist
- [x] All tests still pass.
advance
simple

### DIFF
--- a/eqcorrscan/utils/plotting.py
+++ b/eqcorrscan/utils/plotting.py
@@ -13,7 +13,7 @@ import logging
 import datetime as dt
 import copy
 import os
-import matplotlib.pyplot as plt
+
 import matplotlib.dates as mdates
 from copy import deepcopy
 from collections import Counter
@@ -64,6 +64,7 @@ def _finalise_figure(fig, **kwargs):  # pragma: no cover
     Internal function to wrap up a figure.
     {plotting_kwargs}
     """
+    import matplotlib.pyplot as plt
 
     title = kwargs.get("title")
     show = kwargs.get("show", True)
@@ -173,6 +174,7 @@ def xcorr_plot(template, image, shift=None, cc=None, cc_vec=None, **kwargs):
 
     .. image:: ../../plots/xcorr_plot.png
     """
+    import matplotlib.pyplot as plt
     if cc is None or shift is None:
         if not isinstance(cc_vec, np.ndarray):
             Logger.error('Given cc: %s and shift: %s' % (cc, shift))
@@ -224,6 +226,7 @@ def triple_plot(cccsum, cccsum_hist, trace, threshold, **kwargs):
 
     .. image:: ../../plots/triple_plot.png
     """
+    import matplotlib.pyplot as plt
     if len(cccsum) != len(trace.data):
         Logger.error(
             'cccsum is: ' + str(len(cccsum)) + ' trace is: ' +
@@ -311,6 +314,7 @@ def peaks_plot(data, starttime, samp_rate, peaks=None, **kwargs):
         peaks_plot(data=data, starttime=UTCDateTime("2008001"),
                    samp_rate=10, peaks=peaks)
     """
+    import matplotlib.pyplot as plt
     peaks = peaks or [(0, 0)]
     npts = len(data)
     t = np.arange(npts, dtype=np.float32) / (samp_rate * 3600)
@@ -425,6 +429,7 @@ def cumulative_detections(dates=None, template_names=None, detections=None,
                               rate=True, show=True)
 
     """
+    import matplotlib.pyplot as plt
     from eqcorrscan.core.match_filter import Detection
     # Set up a default series of parameters for lines
     colors = cycle(['red', 'green', 'blue', 'cyan', 'magenta', 'yellow',
@@ -589,6 +594,7 @@ def threeD_gridplot(nodes, **kwargs):
         threeD_gridplot(nodes=nodes)
     """
     from mpl_toolkits.mplot3d import Axes3D  # noqa: F401
+    import matplotlib.pyplot as plt
     lats = []
     longs = []
     depths = []
@@ -687,6 +693,7 @@ def multi_event_singlechan(streams, catalog, station, channel,
 
     .. image:: ../../plots/multi_event_singlechan.png
     """
+    import matplotlib.pyplot as plt
     # Work out how many picks we should have...
     short_cat = Catalog()
     short_streams = []
@@ -796,6 +803,7 @@ def multi_trace_plot(traces, corr=True, stack='linstack', **kwargs):
          stack: 'linstack' or 'PWS', or None.
     {plotting_kwargs}
     """
+    import matplotlib.pyplot as plt
     from eqcorrscan.core.match_filter import normxcorr2
     n_axes = len(traces)
     if stack in ['linstack', 'PWS']:
@@ -922,6 +930,7 @@ def detection_multiplot(stream, template, times, streamcolour='k',
                             times=times)
 
     """
+    import matplotlib.pyplot as plt
     # Only take traces that match in both accounting for streams shorter than
     # templates
     template_stachans = [(tr.stats.station, tr.stats.channel)
@@ -1041,6 +1050,7 @@ def interev_mag(times, mags, **kwargs):
         times = [event.preferred_origin().time for event in catalog]
         interev_mag(times, magnitudes)
     """
+    import matplotlib.pyplot as plt
     info = [(times[i], mags[i]) for i in range(len(times))]
     info.sort(key=lambda tup: tup[0])
     times = [x[0] for x in info]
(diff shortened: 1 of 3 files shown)


calum-chamberlain:
Can you put the matplotlib imports back as they were please.  They need to be like like for stable imports and backend management.

imankahbasi:
> Can you put the matplotlib imports back as they were please. They need to be like like for stable imports and backend management.

Hi,
Yes, i put them back.

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+        the origin time of each template in match_filter or being changed to
+        origin time later.
+        Something like 2014-10-18T04:02:02
+        If template's name is something else, only detections will be plot.
+
+    .. rubric:: Example
+    >>> from eqcorrscan.core.match_filter import read_detections
+    >>> from eqcorrscan.utils.plotting import origintime_detections
+    >>> # Read text file detections
+    >>> detections = read_detections('path/to/file/filename')
```
calum-chamberlain: This fails because this file doesn't exist.

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+        the origin time of each template in all detections.
+        Something like 2014-10-18T04:02:02
+        '''
+    import matplotlib.pyplot as plt
+    day = UTCDateTime(detections[0].detect_time.date)
+    times = [day.datetime, (day+(3600*24)).datetime]
+    plt.figure()
+    plt.plot(times, [0, 0])
+    if method == 'simple':
+        xdetects = [detect.detect_time.datetime for detect in detections]
```
calum-chamberlain: It looks like you are plotting detection-time, not origin-time, but your xlabel - origin time?

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+        try:
+            parents = [detection.template_name for detection in detections]
+            parents = list(set(parents))
+            number_parents = len(parents)
+            parents = [UTCDateTime(parent).datetime for parent in parents]
+            xparents = [parent for parent in parents if
+                        parent.date() == day.date]
+            yparents = np.zeros(len(parents))
+            plt.plot(xparents, yparents, '.', color='blue', label='parents')
+        except Exception as exception:
```
calum-chamberlain: Can you be more specific in your caught exception please?  Also, can you wrap only the statement that you expect to generate an expection in a try.. except rather than the whole thing.  It is generally bad practice to wrap with a plain `Exception`, and to wrap all the code.  In general, avoiding `try.. except` and using a well written `if` statement is better as well.

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+            print(exception)
+            print(txt)
+    elif method == 'advance':
```
calum-chamberlain: Do you mean "advanced" rather than "advance"?

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+            print(txt)
+    if number_parents > 30:
+        ncol = 2
+    else:
+        ncol = 1
+    legend = plt.legend(loc=(1.04, 0), edgecolor='g', ncol=ncol,
+                        title='parents({})\nchild-events({})'
+                        .format(number_parents, len(detections)))
+    legend._legend_box.align = "left"
+    plt.xlabel('Origin time')
```
calum-chamberlain: What is your y-axis?

eqcorrscan/utils/plotting.py
```diff
@@ -2184,6 +2184,122 @@ def subspace_fc_plot(detector, stachans, **kwargs):
...
+    plt.gca().axes.get_yaxis().set_visible(False)
+    fig = plt.gcf()
```
calum-chamberlain: Can you re-write this so that a user can pass a figure  as an argument please.

imankahbasi:
> I'm not really sure what the use-case is for this, and why it should be included as an included plotting function in EQcorrscan. Can you justify why we should include this and continue maintaining it please?

In my opinion, visual output is useful for understanding.
In this plot we can find how many events are detected by each template, which one detects more and which one detects less, density of detection in time and etc in light, fast and simple plot. Maybe we can find our mistakes and solve them. It's interesting to compare detection in cccs plot and text output of match_filter.
It will be good if all examples in EQcorrscan tutorial have visual output, like obspy tutorial.
And EQcorrscan will include more function for visual check of output.

eqcorrscan/utils/plotting.py
imankahbasi: Yes, i will change xlabel. thanks.

calum-chamberlain:
So is the purpose of this different to the cumulative detection plotting already included?
I agree that plots are helpful, I just don't know where you would call this from, and what different/better information it gives than the cumulative plot, which is supported by `Party.plot()`?

eqcorrscan/utils/plotting.py
imankahbasi: Yes, i will write it for ValueError exception.
I am interesting to write it with if statement, but i don't know how can i find a string is proper for UTCDateTime or not yet.

eqcorrscan/utils/plotting.py
imankahbasi: I don't understand what kind of change you would like. can you explain more?

eqcorrscan/utils/plotting.py
imankahbasi: > So is the purpose of this different to the cumulative detection plotting already included?
> I agree that plots are helpful, I just don't know where you would call this from, and what different/better information it gives than the cumulative plot, which is supported by `Party.plot()`?

You right. I'm not sure. Maybe this function is same as cumulative detection plot or even less usefull. Maybe it is another aspect of this mean.
If I'm being honest, in my case i had a simple mistake in my code then i detect duplicate of one event. after i saw this graph i found i had 2 events but only one point in figure.
advance
true


I also noticed an issue in v0.3.3 that I don't know if there is in the new version or not.
I make a template of a specific day, then i use the template for this day to detect event.
but i did not see any detection or even self detection in cccs plot in match_filter function!
but after i draw my plot, i found there is self detection in detection and even more!
true
cccsum_plot_2014-10-10T10_49_11_2014-10-10

imankahbasi:
> So is the purpose of this different to the cumulative detection plotting already included?
> I agree that plots are helpful, I just don't know where you would call this from, and what different/better information it gives than the cumulative plot, which is supported by `Party.plot()`?

You right. I'm not sure. Maybe this function is same as cumulative detection plot or even less useful. Maybe it is another aspect of this mean.
If I'm being honest, in my case i had a simple mistake in my code then i detect duplicate of one event.
false

after i saw this graph i found i had 2 events but only one point in figure.
advance

I also noticed an issue in v0.3.3 that I don't know if there is in the new version or not.
I make a template of a specific day, then i use the template for this day to detect event.
but i did not see any detection or even self detection in cccs plot in match_filter function!
cccsum_plot_2014-10-10T10_49_11_2014-10-10

but after i draw my plot, i found there is self detection in detection and even more!
advance

calum-chamberlain:
That is strange that the detection spikes were not showing up in the matched-filter plot - are you sure the detection plot is for one of the templates that was on that day?

I think the scope of this plot is very similar to the cumulative detection plots, but, those plots do not show the template time, which would be very helpful.  I suggest that you work on adding template times into that plot, with your vertical lines as here.  Personally I find the little crosses close to one another difficult to see, whereas an increasing cumulative count with time I find easier to read.

imankahbasi:
>That is strange that the detection spikes were not showing up in the matched-filter plot - are you sure the detection plot is for one of the templates that was on that day?

I saw this in some case, i will check more for better explanation with some example and details and i'll tell you later.

> I think the scope of this plot is very similar to the cumulative detection plots, but, those plots do not show the template time, which would be very helpful. I suggest that you work on adding template times into that plot, with your vertical lines as here. Personally I find the little crosses close to one another difficult to see, whereas an increasing cumulative count with time I find easier to read.

Yes, although i use this plot in interacting mode and zoom on for better view, then i didn't notice crosses close together. You are right, if i save output of this graph in an image, it's useless. specially in case of numerous ditections, except legend of plot.
Ok, it's good suggestion. i will work on cumulative detection plot and add some feature to it.
Although i like origintime_detections  in some cases for myself :)
do you want to close this PR?

calum-chamberlain:
Yup, I would go for closing this and opening a new PR for the cumulative plot.

PR #337, opened 2019-10-10
WIP: Better multi core

calum-chamberlain:
### What does this PR do?
1. Turns outer-core threading back on (this appears to be stable on my machines)
2. Uses larger FFT lengths than previously set as default - this gives better performance over a wider range.  Users still have the option to change the fft length.

- Large multi-core machines (>40 threads) were not being well-loaded by version 0.4.0.

### PR Checklist
- [ ] All tests still pass.


calum-chamberlain:
Some timings from my 12-thread home machine running this set up:
```python
import numpy as np
from obspy import read
from eqcorrscan.utils.correlate import _fftw_stream_xcorr

st = read()
for tr in st:
    tr.data = np.random.randn(8640000)
template = st.copy()
for tr in template:
    tr.data = np.random.randn(300)

templates = [template for _ in range(100)]

# Varying cores and cores_outer
%timeit _fftw_stream_xcorr(templates=templates, stream=st, cores=12, cores_outer=1)
```
*Note, only 3 channels used, so max of outer cores is 3*
- **Inner: 12, Outer: 1**: `15.1 s ± 445 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
- **Inner: 6, Outer: 2**: `17.1 s ± 272 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
- **Inner: 4, Outer: 3**: `14.7 s ± 392 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

PR #364, opened 2019-12-01
See what github actions is about

calum-chamberlain:
### What does this PR do?

Tries using github-actions for testing.

Just interested to see how it works.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,0 +1,27 @@
+name: Python package
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


d-chambers:
Have any luck? I have really liked GH actions but there are still some features to be desired. It is new though, maybe it will get better.

calum-chamberlain:
No, but I didn't really try.  Something to come back to in the future maybe.  I like the idea of integrating more things into github, just for less fluff, but I am quite fond of circleci at the moment.  I was really keen to spend more time using gitlab, mostly because of their CI integration, and ability to run CI stuff on another, user-owned, server (like one with a GPU...), but haven't explored enough yet.

So many services, so little time.

Issue #376, opened 2020-04-06
Incorrect filter design for filter gain removal in mag_calc.py


twinder36:
**When correcting for gain of pre_filt in mag_calc, a digital filter design is used rather than analog, which is not appropriate to look up the filter gain for a given frequency.**

In ObsPy the iirfilter designed to apply a butterworth bandpass filter is digital, so that it is suitable to be applied using scipy.signal.sosfilt().

To look up the filter gain for a given frequency (and correct the amplitude observation for this) an analog filter should be used (as expected by the ObsPy response functions).

Current code: 
```
z, p, k = iirfilter(corners, [lowcut / (0.5 * tr.stats.sampling_rate),
                           highcut / (0.5 * tr.stats.sampling_rate)],
                           btype='band', ftype='butter', output='zpk')
filt_paz = {'poles': list(p), 'zeros': list(z), 'gain': k,
                  'sensitivity': 1.0}
amplitude /= (paz_2_amplitude_value_of_freq_resp(
                    filt_paz, 1 / period) * filt_paz['sensitivity'])
```

Correct filter design:
```
z, p, k = iirfilter(corners, [lowcut * 2 * np.pi, highcut * 2 * np.pi],    # For analog filter, Wn in rad/s
                          btype='band', ftype='butter', output='zpk', analog=True)
```

See https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.iirfilter.html

calum-chamberlain:
Morning, thanks for looking at this. It looks like you are correct - apologies for this, I clearly didn't look into `paz_2_amplitude_value_of_freq_resp` properly. 
Is the analog equivalent actually equivalent, or should we be correcting using something like:
```python
from scipy.signal import iirfilter, sosfreqz

sos = iirfilter(corners, [lowcut / (0.5 * tr.stats.sampling_rate),
                                   highcut / (0.5 * tr.stats.sampling_rate)],
                     btype='band', ftype='butter', output='sos')
_, gain = sosfreqz(sos, worN=[1 / period], fs=tr.stats.sampling_rate)
amplitude *= gain
```

calum-chamberlain:
@twinder36 I had a go at fixing this using the method I suggested above using the digital filter in the PR here: #378 

It would be great to get your feedback on that.

twinder36:
[...]

PR #378, opened 2020-04-07
Patch amp pick event

calum-chamberlain:
### What does this PR do?
Fixes the bug found in #376.  This PR adds a simple synthetic test for accuracy for filtered an unfiltered data. This test has quite broad error bounds highlighting the inaccuracy of the method as written.

I also took the opportunity to rewrite the function to make the code less repetitive and cleaner. It still isn't perfect, but its better than the old cludge.

#376 highlighted that I was looking up the wrong gain by assuming that the lookup method used previously would get the correct gain for the digital filter, when in actual fact it is implemented for analogue filters.  This resulted in wildly inaccurate gains computed (often > 100x).  This was the source of an issue patched in development when amplitude were divided by 100. This hack has been removed.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,2 +1,3 @@
 linters:
-  flake8: {  }
+  flake8:
+    python: 3
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
+## Current
+* utils.mag_calc:
+  - `amp_pick_event` now works on a copy of the data by default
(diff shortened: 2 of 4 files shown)


twinder36:
Implementation looks good, but the comments on mag_calc.py lines 889-891 should be updated to reflect using sosfreqz rather than the analog filter paz

twinder36:
You will also need np.abs(gain[0]) else you will return an array

calum-chamberlain:
Network timeouts - merging

PR #390, opened 2020-04-28
Suppress spike_test errors and auto-clean data

imankahbasi:
### What does this PR do?
Spike test function is not make errors anymore.
With removing the spiky data the process will continue on the rest of data.
### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,7 @@
 ## Current
+* Better approch to handling spike.
+  - Spike test function is not make errors anymore.
+    With removing the spiky data the process will continue on the rest of data.
 * Add seed-ids to the _spike_test's message.
 * utils.correlation
   - Cross-correlation normalisation errors no-longer raise an error
--- a/eqcorrscan/core/match_filter/helpers.py
+++ b/eqcorrscan/core/match_filter/helpers.py
@@ -45,7 +45,6 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
     :param multiplier: Multiplier of range to define a spike.
     :type multiplier: float
     """
-    from eqcorrscan.core.match_filter.matched_filter import MatchFilterError
 
     list_ids = []
     for tr in stream:
@@ -62,7 +61,10 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
                'FFT prior to cross-correlation.\n' +
                'If you think this spike is real please report ' +
                'this as a bug.')
-        raise MatchFilterError(msg)
+        print(msg)
+        for ID in list_ids:
+            stream.remove(stream.select(id=ID)[0])
+            print('%s got removed by EQcorrscan because it had spike' % ID)
 
 
 def _total_microsec(t1, t2):


CHANGES.md
```diff
@@ -1,4 +1,7 @@
 ## Current
+* Better approch to handling spike.
```
calum-chamberlain, line 2: I wouldn't say this is a better approach - just different.
```suggestion
* spike_test: Errors changed to warnings when spikes found, spiked data removed and correlations continue
```

eqcorrscan/core/match_filter/helpers.py
```diff
@@ -62,7 +61,10 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
                'FFT prior to cross-correlation.\n' +
                'If you think this spike is real please report ' +
                'this as a bug.')
-        raise MatchFilterError(msg)
+        print(msg)
```
calum-chamberlain, line 64: EQcorrscan doesn't use print statements.
```suggestion
        Logger.warning(msg)
```

eqcorrscan/core/match_filter/helpers.py
```diff
@@ -62,7 +61,10 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
...
+        for ID in list_ids:
+            stream.remove(stream.select(id=ID)[0])
+            print('%s got removed by EQcorrscan because it had spike' % ID)
```
calum-chamberlain, line 67:
```suggestion
            Logger.info(f"{_id} was removed by EQcorrscan because it had spikes")
```

eqcorrscan/core/match_filter/helpers.py
```diff
@@ -62,7 +61,10 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
                'FFT prior to cross-correlation.\n' +
                'If you think this spike is real please report ' +
                'this as a bug.')
-        raise MatchFilterError(msg)
+        print(msg)
+        for ID in list_ids:
```
calum-chamberlain, line 65: variables should not use capitals.
```suggestion
        for _id in list_ids:
```

eqcorrscan/core/match_filter/helpers.py
```diff
@@ -62,7 +61,10 @@ def _spike_test(stream, percent=0.99, multiplier=1e7):
...
+            stream.remove(stream.select(id=ID)[0])
```
calum-chamberlain, line 66: The reason I *don't* do this is because it removes a lot of data for potentially just one spiked data point. This seems like "throwing the baby out with the bathwater". I leave it up to the user to work out what to do about the spikes found by this function. In general removing the whole trace seems like the worst option. If you want to pursue this I would recommend just cutting-out a region in the trace around the spike. The pre-processing functions cope with gaps in data well so this could be a good option.

calum-chamberlain:
@imankahbasi I think solving this (e.g. making spike-test accurately pick up on things that will cause an issue with cross-correlation, and warning the user and optionally removing those data points) is really worthwhile, but I don't think this is the right way to do it for the reasons above - I'm going to close this PR.

I'm thinking about writing a C-function for spike-testing that effectively does the cross-correlation with one dummy template with all channels, and returns a mask to Python that can then be checked and issue locations properly reported. Unfortunately I don't have time to do this at the moment.

Issue #388, opened 2020-04-27
Check full SEED id of picks against Trace header information


cjhopp:
**Describe the bug**
Currently, picks used when generating a Template are only checked against the Station and Channel attributes in the Trace header. For cases ([this station](https://ds.iris.edu/mda/NV/NSMTC/?starttime=2017-08-18&endtime=2599-12-31), for instance) where multiple sensors exist at a given set of lat/lon coordinates (thereby requiring use of the underappreciated SEED location code), multiple picks with the same `phase_hint` will fulfill the following conditions:

```
tr.stats.station == pk.waveform_id.station_code and
tr.stats.channel == pk.waveform_id.channel_code
```

**Expected behavior**
We instead want only the picks on a single sensor, where the sensors are distinguished only by their location code.

Ditto Network information, which likely produces a similar result if you happen to find stations of the same name but in different networks.

**Desktop (please complete the following information):**
 - Operating System: Arch
 - Python version: 3.7
 - EQcorrscan version: 0.4.0

calum-chamberlain:
Definitely a "feature" not a bug :stuck_out_tongue_closed_eyes: 

But yes, would be great to check the full seed-id.

As stated on gitter, a key here will be in doing this change "gently": change with optional opt-in for next minor release with a warning that opt-in will be ~mandatory~ *the default* (with optional opt-out) in the next major release.  Then when we break everyone's codes in the next major release we can at least say:
> we told you so...

flixha:
I had the impression that for several other steps where the location and network codes can be an issue, one has to manually modify them to a common name anyways (e.g., for correlating templates and continuous traces). But I guess one reason that network+location code may not be checked for the Picks is that in Nordic format (from Seisan), such information is not stored anyways - because it usually doesn't matter for the pick (of course it could if one looks at sensors that are a few 100 m apart, and earrthquakes close by are the target).
It's good to keep in mind all kinds of changes that can happen to location codes, e.g., changes through the lifetime of a station, differences between data repositories for the same network (EIDA vs IRIS), and differences between different transmission systems for the same network from the same data provider (e.g., FTP vs FDSNWS). Hence I feel there could also be a use of allowing an option to respect or ignore location+network codes globally across the code. What do you think?

cjhopp:
Hi @flixha!

The approach @calum-chamberlain outlines above would (I think) allow you to opt out of network+location checks even after the next minor release, but would change the default behavior to opt-in.

[...]

PR #391, opened 2020-04-28
Check full seed

cjhopp:
### What does this PR do?
Uses full SEED information (adds network and location) when comparing e.g. pick.waveform_id and tr.stats in a number of functions. Mostly importantly template_gen, but may also include mag_calc and others. 

 #388 

In situations where the desired template or trace contains unique picks with the same Station and Channel, but different location and/or Network codes (e.g. separate sensors in one borehole), current behavior does not differentiate between them.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/eqcorrscan/core/template_gen.py
+++ b/eqcorrscan/core/template_gen.py
@@ -56,7 +56,7 @@ def template_gen(method, lowcut, highcut, samp_rate, filt_order,
                  all_horiz=False, delayed=True, plot=False, plotdir=None,
                  return_event=False, min_snr=None, parallel=False,
                  num_cores=False, save_progress=False, skip_short_chans=False,
-                 **kwargs):
+                 check_full_seed=False, **kwargs):
     """
     Generate processed and cut waveforms for use as templates.
 
@@ -120,6 +120,13 @@ def template_gen(method, lowcut, highcut, samp_rate, filt_order,
         Whether to ignore channels that have insufficient length data or not.
         Useful when the quality of data is not known, e.g. when downloading
         old, possibly triggered data from a datacentre
+    :type check_full_seed: bool
+    :param check_full_seed:
+        If True, will check the trace header against the full SEED id,
+        including Network, Station, Location and Channel. If False (default),
+        will check only against Station and Channel. This behavior was
+        originally necessary to cope with some software (i.e. SEISAN) not
+        storing picks with full SEED info.
 
     :returns: List of :class:`obspy.core.stream.Stream` Templates
     :rtype: list
@@ -394,7 +401,7 @@ def template_gen(method, lowcut, highcut, samp_rate, filt_order,
             template = _template_gen(
                 event.picks, st, length, swin, prepick=prepick, plot=plot,
                 all_horiz=all_horiz, delayed=delayed, min_snr=min_snr,
-                plotdir=plotdir)
+                plotdir=plotdir, check_full_seed=check_full_seed)
             process_lengths.append(len(st[0].data) / samp_rate)
             temp_list.append(template)
             catalog_out += event
--- a/eqcorrscan/core/template_gen.py
+++ b/eqcorrscan/core/template_gen.py
@@ -589,7 +589,7 @@ def _rms(array):
 
 def _template_gen(picks, st, length, swin='all', prepick=0.05,
                   all_horiz=False, delayed=True, plot=False, min_snr=None,
-                  plotdir=None):
+                  plotdir=None, check_full_seed=False):
     """
     Master function to generate a multiplexed template for a single event.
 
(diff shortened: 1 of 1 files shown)


calum-chamberlain:
From your messages the other day this is still a work in progress? If so, can you add "WIP: " to the start of this PR title until you are ready for review?

calum-chamberlain:
When you are ready you can update this - I fixed some test issues on master and merged that into develop - it should help with CI things!

cjhopp:
@calum-chamberlain 

Changing _make_event_pair() to force identical seed id's, same as the correlations. I can see this being a problem if we want to allow differential times for S-picks even if on different channels at the same sensor. Thoughts?

As it is, its producing problems with my silly borehole sensors (same station, different location codes) so that my dt.ct contains bogus times whereas my dt.cc is fine. But that can also fall on me to rename them as @flixha had mentioned previously.

calum-chamberlain:
Changing to full seed-id checking for `_make_event_pair` could be a little annoying for me - but it is probably the right way to do it! Can you allow an option to just check the station as well though? I imagine there would be some instances where people (including me) only care about the station and not the full seed id.

cjhopp:
Yeah, mine is obviously the edge case!

The default, for now, will be checking station only, as discussed in #388

calum-chamberlain:
Looks like the history of this one got pretty ugly!

calum-chamberlain:
Changing the names would be simpler, but it is definitely not the "correct" way to do it - the only reason I didn't have full seed ID checking in from the start was because SEISAN didn't keep the full seed ID...  I'm keen to get this "corrected" and suck up the issues I create for myself because it will be "better" - as long as well actually produce useful errors when things go wrong. I need to patch some catalog-to-dd things and include more useful errors there. I have spent too much time working out what I have done wrong.

No rush, I think this one should be a "major" release change, so I'm not in a hurry to get it out for 0.4.3 with will contain a lot of bug-fixes from @flixha.

flixha:
> Just how I like my PR's.
> 
> I'm okay blowing this up, if you'd like to clean house. I obviously haven't had a chance to work on this in forever and a day...
> 
> It'd be useful to have, but the more it touches, the more I think @flixha was right and that maybe renaming sensors with different locations at the same station is the better approach?

The more networks and data I see the more I understand the challenges and that user needs some freedom in the approach. Maybe it would be nice to have two supported approaches, one which completely ignores network and location codes, and one that takes them fully into account. I can't fully see yet how hard that would be and in how many places in the code that would need to be considered, but it could help to solve a good set of common problems that come up related to the issue:

- as you mentioned, multiple sensors at different levels in a borehole (with different location codes, but same station codes --> need to take all codes into account)
- data populated from databases that only contain information the station and channel (old Nordic format; lots of ISC data --> mostly better to ignore location & network codes)
- data from networks that have been operating for a long time since when the default handling has changed. That applies to a good set of data of the GSN net, where they often used no location code until the 90s and then changed the very same sensor over to a new location code (--> better ignore location code)

And then there's of course detection problems where all of these challenges arise,  and in those cases the user of course has to make up their own strategy.

calum-chamberlain:
Hey @cjhopp and @flixha - do either of you remember where we are with this one - do you want me to cherry pick changes from this for a clean PR, and add in the option of not checking the full SEED ID? 

I was reminded of this by a user who has the same station code for multiple stations across different networks, and the templates that were being made were not useful!

flixha:
hey @calum-chamberlain , I'm not 100 % sure what the state was as I don't remember testing it properly, and I may have changed my opinion a bit since what I wrote above - e.g., for my own cases I now make sure that I "normalize" all network and location codes explicitly for traces and associated picks that I want to correlate. That leaves a bit more work to the user but is more "correct", along the lines of what @cjhopp implemented here.

To avoid frustrating users who are updating their EQcorrscan with this stricter handling here, it would probably be nice to allow keep the strict SEED-ID handling as an option. However, considering version numbers EQcorrscan is still in its 0.x-stage where such changes can come without early depreciation-warnings etc. So: the option to not check the full SEED ID would be nice, but personally I don't see is as important any more as I did a while ago.

If you're able to cherry-pick this PR that would be great - I can try to look out for any edge cases in some tests soon.

cjhopp:
Hi chaps, 100% agree with @flixha here. My recollection was that I had added a `check_full_seed` arg to most of the necessary places, but never got around to the tests or cleaning it up. As @flixha says above, the consensus was to leave strict SEED checking as an "opt-in" feature so as not to break things for users. @calum-chamberlain , please feel free to cherry-pick whatever you'd like and destroy this thing.

calum-chamberlain:
Awesome, thanks @cjhopp and @flixha - I will try to clean this up today and think about some testing. :) Agree on the optional opt-in and warn that this will change implementation, and the flexibility of 0.x - I don't know if I will ever get this to 1.x! I'm still learning so much about good ways to do things that I always look back on code and want to change it!

PR #402, opened 2020-05-23
The freq_mag function supports 2 groups of magnitudes as children and templates now

imankahbasi:
… templates now to better compare magnitudes before and after detections.

### What does this PR do?
The freq_mag function supports 2 groups of magnitudes as children and templates now to better compare magnitudes before and after detection.
### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 ## Current
+* The freq_mag function supports 2 groups of magnitudes as children and
+  templates now to better compare magnitudes before and after detections.
 * Add seed-ids to the _spike_test's message.
 * utils.correlation
   - Cross-correlation normalisation errors no-longer raise an error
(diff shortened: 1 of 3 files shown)


calum-chamberlain:
I'm not that keen on this - mostly for changing the API without much benefit. If you want to persue this I would suggest going down the route of keeping the same function signature, but allowing magnitudes to be a an `iterable` (as it is now), or an `iterable` of `iterable`s, and the other arguments to be either single values (as now) or `iterable`s.  Then you could pass as many magnitude sets as you like.

In general changing argument names is best avoided - if it is unavoidable then we have to make a new major release and justify to users that they frustration at their codes breaking is with good reason.

calum-chamberlain:
Closing due to lack of response

PR #419, opened 2020-07-14
Writing out detections: add 'SDS' as archive-option, plus bug-fix with party-overwrite

flixha:
### What does this PR do?
1. For `party.write`: 
  -- Fixes a bug with overwriting of `.tgz`-files
  -- Adds the option `overwrite` (default: False) so the user can manually choose the previous (though unintended) option, which makes rerunning the lag-calc-picker during tests a bit easier.
2. For `eqcorrscan/utils/archive_read.py`:
-- Adds the option `arctype='SDS'`to read from a SeisComp Data Structure (SDS) archive.
-- Adds `SDS`-option to ` _check_available_data`.
-- Adds support to request data with wildcards.

### Why was it initiated?  
1. For `party.write`: 
There was a bug that if `format=tar` is  chosen and a filename without `.tgz`-ending is supplied, then `party.write` overwrote the `.tgz`-file, even though the function was previously intended to never overwrite any existing party-files.
2. For `eqcorrscan/utils/archive_read.py`:
-- Previously, the supported archive-types were `seishub`, `FDSN`, `day_volumes`, but `SDS` was not supported yet. I wanted to use  `eqcorrscan.utils.clustering.extract_detections` to extract detection-waveforms from an SDS-archive.
-- Wildcard-requests were already being used to work around 2-vs-3-letter channel-name issues, but a comparison with a wildcard never passed the following condition:
https://github.com/eqcorrscan/EQcorrscan/blob/c1512d7b3cff627556850c6420f59406ed009b09/eqcorrscan/utils/archive_read.py#L109

### Any relevant Issues?
For the `archive_read(arctype='SDS')`, there will need to be some tests (see comments to this PR for more).

### PR Checklist
- [X] All tests still pass.


flixha:
Regarding tests for `arctype='SDS'`: I can create a minimum-example SDS-archive and commit these files in `eqcorrscan/tests/test_data`. 
Until now I have only used the `SDS`-option when using `eqcorrscan.utils.clustering.extract_detections`, as I'm not sure where else this option should possibly be added (e.g., for `template_gen`). For the test, it should probably be tested from the highest level of support in the functions (in comparison, I see that `arctype=seishub` is tested in `tests/template_gen_test.py`

flixha:
I see that the tests are passing on macOS, but not on Ubuntu now - now that was unexpected ;-) !

eqcorrscan/core/match_filter/party.py
```diff
@@ -632,6 +632,10 @@ def write(self, filename, format='tar', write_detection_catalog=True,
             SC3ML, QUAKEML are supported. Note that not all information is
             written for all formats (QUAKEML is the most complete, but is
             slow for IO).
+        :type overwrite: boolean
```
calum-chamberlain, line 635:
```suggestion
        :type overwrite: bool
```

eqcorrscan/utils/archive_read.py
```diff
@@ -16,11 +16,13 @@
 import os
 import logging
 import glob
+import fnmatch
```
calum-chamberlain, line 19: Nice, I didn't know about this module before, thanks for the education!

eqcorrscan/utils/archive_read.py
```diff
@@ -216,6 +231,11 @@ def _check_available_data(archive, arc_type, day):
                 for channel in station:
                     available_stations.append((station.code,
                                                channel.code))
+    elif arc_type.lower() == 'sds':
```
calum-chamberlain, line 234: Just for consistency: 
```suggestion
    elif arc_type.upper() == "SDS":
```

calum-chamberlain:
Cool, thanks @flixha 

1. Please do create a minimal SDS archive to test against, but can you post a comment here on the size and number of files before you commit that please?  I'm sure it will be less than we already have in test-data.  If there is any-way to re-use the test-data that we already have that would be great...? I'm not sure if that is possible though.
2. I imagine the test-fail on Ubuntu and Windows is coming from `test_family.tgz` being written out by another, concurrent, test... I can't see it though...
3. Can you add a note to `CHANGES.md` please.

flixha:
1. As SDS is also just a folder structure with daylong files, albeit a bit different compared to day_vols, I should be able to just create the right folder structure an symlink the files in `tests/test_data/day_vols`. That would basically not change the size of the test dataset, but of course I'd need to test there's not issue with these symlinks on any of the other test systems.

2. I'll look into it, maybe the option `overwrite=True` can solve it somewhere.

3. Done.

calum-chamberlain (approved):

calum-chamberlain:
Looks like a couple of minor test fails related to party-overwrite.

flixha:
I fixed the failing test, and found that I hadn't added the `overwrite`-option to the `family.write` which I should have done after adding it to `party.write`. Fixed now, and all tests now work for me (except the one that is supposed to fail in this version).

Issue #430, opened 2020-09-20
[HELP]Running Jupyter example from command line


dr-glenn:
# What do you need help with?
I exported the Jupyter notebook example of section 4.1.1. I removed **%matplotlib inline**. Run with Python 3.7.8, same as the Jupyter notebook. Funny things happened. For example, I got four map plots from **fig=catalog.plt**. Processing seemed to stall until I closed each of these figure windows. Then I got four more. This goes on and on until I kill the run.

# Provide an example so that we can reproduce your problem
Program is inline below.

# What help would you like?
Please tell me how this program should be run from command line. First and foremost, why multiple map plots? I know this means somewhere multiple threads are running, but why and how does this happen before any eqcorrscan code?

# What is your setup? (please complete the following information):**
 - Operating System: Windows 10 Pro.
 - Python version: 3.7.8
 - EQcorrscan version: conda install a week ago. 0.4.2.
# Code

```python
    #!/usr/bin/env python
    # coding: utf-8
    # based on eqcorrscan example from 4.1.1
    import logging
    logging.basicConfig(
        level=logging.ERROR,
        format="%(asctime)s\t%(name)s\t%(levelname)s\t%(message)s")
    #get_ipython().run_line_magic('matplotlib', 'inline')
    from obspy import UTCDateTime
    from obspy.clients.fdsn import Client
    from eqcorrscan.utils.catalog_utils import filter_picks
    
    client = Client("NCEDC")
    t1 = UTCDateTime(2004, 9, 28)
    t2 = t1 + 86400
    catalog = client.get_events(
        starttime=t1, endtime=t2, minmagnitude=2.5, minlatitude=35.7, maxlatitude=36.1,
        minlongitude=-120.6, maxlongitude=-120.2, includearrivals=True)
    fig = catalog.plot(projection="local", resolution="h")
    catalog = filter_picks(catalog=catalog, evaluation_mode="manual", top_n_picks=20)
    from eqcorrscan import Tribe
    
    tribe = Tribe().construct(
        method="from_client", lowcut=4.0, highcut=15.0, samp_rate=50.0, length=6.0,
        filt_order=4, prepick=0.5, client_id=client, catalog=catalog, data_pad=20.,
        process_len=21600, min_snr=5.0, parallel=True)
    print(tribe)
    print(tribe[0])
    fig = tribe[0].st.plot(equal_scale=False, size=(800, 600))
[...]

PR #432, opened 2020-10-16
Start an FAQ

calum-chamberlain:
### What does this PR do?

Adds an FAQ page to the EQcorrscan docs.

I frequently get similar questions from users, both via email and through the github page. I think it would be good to:
a) keep a record of these questions to make it easier to field them for the devs
b) provide a place for users to check and problem solve themselves.

There are a few relevant issues (e.g. #430) that should be tagged in the FAQ.

### TODO:

Note, this PR makes no changes to code and is based on master to propagate out quickly. This should not result in a new release.
- [ ] Add links to github issues
- [ ] Add more FAQs @eqcorrscan/maintainer @emilyws1 any suggestions?
- [ ] Fill out the answers and add code snippets where appropriate.
- [ ] Check the rendered docs.


calum-chamberlain:
Docs are rendered [here](https://eqcorrscan.readthedocs.io/en/faq/)

Issue #438, opened 2021-02-12
Should the self-detection time of a template be independent of other templates in the tribe?


flixha:
I ran some tests to let an event template detect itself in continuous data. But depending on the station choice and what other templates were in the tribe that I used, I got different detection times (11 s apart). I tried to debug this and believe I have an idea of what's happening. I'm wondering whether this is the expected behaviour or not. If it's not expected, it would be nice if you have some feedback for how to solve it.

### Issue description:
- Let's detect an event with a template of itself in continuous data. There are two templates, A and B, with template A being created from the event that we want to detect itself.
  - the earliest trace (i.e., pick) in the templates is for station DRUM, but for that station there is less than 80 % data during that day
  - template B contains a pick + trace for station INVG, but there is not such pick or trace for template A
  - both templates contain pick+trace for stations EKB10.BHZ and ESK.BHZ, but template B contains EKB10.BHZ twice because it's a 1-comp station with P and S-pick on the same channel

- In case 1
  - I use a tribe of two templates, A and B , with template A created from the event itself.
  - for template A, `eqcorrscan.utils.preprocessing._prep_data_for_correlation` removes the traces for DRUM and adds two NaN-traces (for EKB10.BHZ and INVG.BHZ, with both traces starting at 18:14:42
  - the stream contains traces EKB10.BHZ x2, ESK.BHZ, INVG.BHZ
  - the event is detected at **18:14:42** (hh:mm:ss, same a P-pick on DRUM), and the detection-channel list contains [('EKB10', 'BHZ'), ('ESK', 'BHE'), ('ESK', 'BHN'), ('ESK', 'BHZ')]
- In case 2
  -  I use a tribe of one template, A, created from the event itself.
  - for template A `eqcorrscan.utils.preprocessing._prep_data_for_correlation` removes the trace for DRUM and INVG, so that the earliest remaining trace starts at 18:14:53
  - the stream contains traces EKB10.BHZ, ESK.BHZ
  - the event is detected at **18:14:53** (hh:mm:ss, i.e., P-pick on EKB10), and the detection-channel list mentions [('EKB10', 'BHZ'), ('ESK', 'BHE'), ('ESK', 'BHN'), ('ESK', 'BHZ')]

### Analysis / Consequences:
[...]

PR #439, opened 2021-02-15
Fix bug where detection time depends on content of a tribe when continuous data are incomplete

flixha:
In reference to issue #438:

### What does this PR do?
- add test to check for the bug (bug is described below)
- WIP: will add details once I have fixed the bug

    then don't remove it, but instead add a NaN-channel for that station-channel to the stream

### PR Checklist
- [ ] All tests still pass.


flixha:
- Right now, the tests are supposed to fail with the the new test checking for the bug (they do).
- However, the reason they are failing in the test environments right now are some problems with the conda-installation in the tests.
- In my local tests, I also had issues with the NCEDC-fdsn server that is queried in the match-filter tests: For the selected time span, the data were extremely slow to download (~2 KB/s), so that the client timed out before the download was complete.

calum-chamberlain:
Thanks for this.  However can you try to clean up the commit history and makes the changes for this pr in a clean develop branch?

I'm going into the field tomorrow and won't get to look at this until next week,  sorry.

CJ Chamberlain, out of office

flixha:
Yes I'll clean that up. Right now I'm thinking about how to implement the checks in the most resource-saving way. What I'm thinking is that we only need to add NaN-traces to the stream in one specific case: if there is a template for which continuous data for the earliest trace are missing, and which at same time has not been filled with any NaN-traces (because NaN-traces in the template are set to the starttime of the earliest trace). So this is a rather specific case, and if we add the NaN-trace to the stream only in that specific case, then the overall extra resource usage should be very limited.

flixha:
I have added some code  to `eqcorrscan.utils.preprocessing._prep_data_for_correlation` now that will add a NaN-trace into the continuous stream and the templates if that is required to work around this bug. The detection now gets the correct `detect_time` and the correct channel list.

For the new test, I now avoid checking the events. As `compare_families` checks the channel-list in the detection against the original channels in the template, this test would always fail if we make the continuous data gappy or missing.

calum-chamberlain:
Thank @flixha - I'm hoping to get onto reviewing this on Friday, and fixing other failing tests. Sorry fixing the actions startup fails took so long!

flixha:
No worries @calum-chamberlain ! I wish I could have solved the actions myself, but as I haven't worked with these myself I figured you would be much quicker at that (still time for me to have a proper look I guess). I had all tests passing locally on RHEL, so I hope there's nothing major that bothers the tests on Ubuntu...

eqcorrscan/utils/pre_processing.py
```diff
@@ -846,10 +846,39 @@ def _prep_data_for_correlation(stream, templates, template_names=None,
     # Remove templates with no matching channels
     filt = np.ones(len(template_names)).astype(bool)
     for i, template in enumerate(templates):
-        template_ids = {tr.id for tr in template}
-        if len(template_ids.intersection(stream_ids)) == 0:
+        trace_ids = {tr.id for tr in template}
+        if len(trace_ids.intersection(stream_ids)) == 0:
```
calum-chamberlain, line 850: It might be a good idea/necessary to apply this filter before the addition of empty channels for early data. This `filt` should be used to remove templates that have nothing in common with the stream, which we still want to do regardless of this issue. Removing these before adding empty channels will at best make this faster (fewer templates to check), and potentially get around any bugs whereby nan channels are added to the stream for these templates that have nothing in common with the stream.

eqcorrscan/utils/pre_processing.py
```diff
@@ -846,10 +846,39 @@ def _prep_data_for_correlation(stream, templates, template_names=None,
...
+    # Ensure that the templates' earliest traces are kept, even if there is no
+    # continuous data for them. If this happens, we need to add a NaN-stream to
+    # the continuous data to avoid inconsistent detection times.
+    n_template_traces = np.array([len(temp) for temp in templates])
+    n_stream_traces = sum([n+1 for s, n in seed_ids])
+    # These checks are not necessary if all templates will get NaN-traces,
+    # because the NaN-traces will save the right starttime for the template.
+    nan_stream_ids = list()
+    if any(n_template_traces > n_stream_traces):
+        earliest_templ_trace_ids = list(set(
```
calum-chamberlain, line 862: May not need to convert from set to list.

calum-chamberlain (commented):
Couple of minor comments - once those are addressed, if this fixes the bug then this is golden. Thank you!

eqcorrscan/utils/pre_processing.py
flixha: True; don't need to convert to list.

eqcorrscan/utils/pre_processing.py
flixha: True; I moved the filtering to before any NaN-traces are added to the stream, and made sure all tests are carried out on the filtered set of templates.

flixha:
Thanks for getting on all these PRs at once! I addressed the two comments that you had above, thanks for catching the implications that the wrong order may have had. I committed the fixes now. 

Any idea where the test fails for  `clustering_test.py` are coming from? I have tested this branch on RHEL and MacOS, and there was no issue with the clustering-test (or other tests)...

calum-chamberlain:
I can reproduce the clustering test issue - there definitely is one in this branch related to the change in shape of correlation outputs, however, fixing that (changing `_positions = np.empty((n_streams, max(no_chans)))` to `_positions = np.empty_like(positions)` brings up another error, with correlations not being the same forward and backward for clustering.

Unfortunately the rest of my week is busy, but this does need fixing. I suspect this is a separate bug that we have exposed here, but it would be good to fix it here.

calum-chamberlain:
I think the issue relates to the 2D FFT used for frequency domain correlations - changing the order between different "core" events in the clustering results in a slightly (10<sup>-5</sup>) different correlation for different channels. This has not been enough to worry about in other applications.  I think this raises it's head in this PR because the order of channels is changed by this fix.

I will patch this here in a bit.

calum-chamberlain:
I'm leaning towards this being an issue with the channels that are retained by `_prep_for_correlation` now - I get different channels used depending on which stream is used as the "master" stream in the `cross_chan_correlation` function. More on this later.

calum-chamberlain:
After too much head scratching I realised that the data were not being copied before being worked on in place - nan-channels were added to the stream to cope with early template channels not being included, which then stayed as nan channels and were not correlated. I added copying to the `cross_channel_correlation` function and it fixed this locally.

calum-chamberlain (approved):

PR #446, opened 2021-03-23
Allow adjusting minimum CC threshold for lag-calc picking individually for detections 

flixha:
### What does this PR do?
This PR implements an extra option in `lag_calc` (`party.lag_calc`, `family.lag_calc`, and  `lag_calc.xcorr_pick_family`) which allows the user to set the minimum CC threshold for a pick to be set in a variable way.  Now the user can choose to set the minimum CC threshold according to `min(detect_val / n_chans * min_cc_from_mean_cc_factor, min_cc)`, i.e., `min_cc` then becomes an upper bound to the minimum CC threshold, while the lower bound is computed from the average CC of each detection and a multiplication factor. By setting `min_cc_from_mean_cc_factor=1`, the user can for example achieve that only traces that correlate better than the detection's average CC return a pick.

One of the most flexible parameters to recognize relevant detections in EQcorrscan is the MAD threshold, where the user defines a factor by how much the MAD has to be exceeded to set a detection.  One can achieve a low rate of misdetection by setting the MAD high enough, e.g., to 10, 15, or even higher depending on the dataset. The user can achieve better separation of correct detections and misdetections by using more stations, longer traces, and a higher frequency cutoff. But while all of these parameters increase the MAD exceedance, they tend to decrease the single-trace cross correlation, resulting in no picks for a value of `min_cc` (in `lag_calc.xcorr_pick_family`) that commonly appears reasonable. With this PR, the user can now work around the issue of not getting any picks by setting the minimum CC-threshold based on the average CC of each trace in a detection. This allows very clear detections to be picked with the same confidence as before, and at the same time allowing weaker, but robust detections to return a relevant number of picks so that these detections can be located.

- I first implemented this functionality in my user-side scripts, but realized that when using the `party.lag_calc`-syntax, then this functionality needs to be in `lag_calc` rather than on the user side to properly compute the CC threshold for each detection.
- The changes do not affect any previous funtionality, as the default for `min_cc_from_mean_cc_factor` is `None` where it has no effect.
- This PR also fixes an issue with `party.rethreshold`: The function previously returned only detections that have `detect_val > new_threshold`. Especially when using MAD exceedance as an indicator, that did not take into account that the user may like to keep large negative detection values. This is now possible by setting `abs_values=True`, with its default being `False`.

### PR Checklist
- [X] All relevant tests still pass. **(except those that are supposed to fail and some unrelated ones that are related to underlying depreciations. I will address that in another PR)**
- [ ] ~~First time contributors have added your name to `CONTRIBUTORS.md`.~~


flixha:
Some of the failing tests are related to no longer supported syntax in underlying libraries that I have now addressed in #447 together with other depreciated or soon retired function calls.
However, I don't understand why the tests for macOS are shown as successful, while the tests for Ubuntu return the exact same results (7 failed, 354 passed, 3 errors) but are shown as failed...?

calum-chamberlain:
Thanks for this @flixha - I will try and take a look at #447 today, and merge that, then work on #439, then get to this. Sorry for the back-log!

eqcorrscan/tests/match_filter_test.py
```diff
@@ -1242,11 +1242,16 @@ def test_party_rethreshold(self):
...
+        party1 = party.copy().rethreshold(new_threshold=9)
+        for family in party1:
             for d in family:
                 self.assertEqual(d.threshold_input, 9.0)
                 self.assertGreaterEqual(d.detect_val, d.threshold)
+        party2 = party.copy().rethreshold(new_threshold=9, abs_values=True)
+        for family in party2:
+            for d in family:
+                self.assertEqual(d.threshold_input, 9.0)
+                self.assertGreaterEqual(abs(d.detect_val), d.threshold)
```
calum-chamberlain, line 1254: Should check that this actually has some large negative detect_vals that are retained.

calum-chamberlain (commented):
Something that I can check, then I will merge.

eqcorrscan/tests/match_filter_test.py
calum-chamberlain: It looks like there are no negative values retained. I'm going to add a check here, make sure the test fails in the CI, then update the test.

calum-chamberlain (approved):

PR #447, opened 2021-03-23
Syntax / function call updates in preparation of obspy 2.0 (+ h5py, scipy)

flixha:
### What does this PR do?
It fixes some depreciated (related to `scipy`, `eqcorrscan`) and some now no longer supported syntax (related to `obspy 2.0` and `h5py`).

Obspy's `master`-branch contains changes that will come fully into effect with the release of `obspy 2.0`. Some of these changes break current syntax and tests in EQcorrscan.
- As `SDSclient` now has a `read_waveform_bulk`-function, this now tests EQcorrscan's monkey-patching on the `Earthworm.client`
- One change to the tests concerns using `obspy.signal.cross_correlation.correlate` and `obspy.signal.cross_correlation.xcorr_max` instead of the now no longer existant `obspy.signal.cross_correlation.xcorr` ([PR #1979 in Obspy](https://github.com/obspy/obspy/pull/1979)). As a note (discussions would best fit in Obspy repo): If two traces, one with an even and one with uneven number of samples, are correlated, then the new `.xcorr_max` function will return for the index of the maximum correlation an index that ends in .5 instead of an integer. Then the user has to do the rounding to be able to use it as an index in an array.

### PR Checklist
- [X] All tests still pass.
- [X] ~~Any new features or fixed regressions are be covered via new tests.~~
- [X] ~~Any new or changed features have are fully documented.~~
- [X] ~~Significant changes have been added to `CHANGES.md`.~~
- [X] ~~First time contributors have added your name to `CONTRIBUTORS.md`.~~


eqcorrscan/tests/match_filter_test.py
```diff
@@ -408,49 +408,20 @@ def test_catalog_extraction(self):
...
                          template_list=self.templates, st=self.st,
                          threshold=8.0, threshold_type='MAD',
-                         trig_int=6.0, plotvar=False, plotdir='.',
+                         trig_int=6.0, plot=False, plotdir='.',
                          cores=1, extract_detections=True, output_cat=True)
         self.assertEqual(len(detections), 4)
         self.assertEqual(len(detection_streams), len(detections))
         self.assertEqual(len(detection_streams), len(det_cat))
 
-    def test_same_detections_individual_and_parallel(self):
```
calum-chamberlain, line 417: Was there a reason for removing this test?

calum-chamberlain (commented):
Just one question on the removal of one test, otherwise, thanks!

eqcorrscan/tests/match_filter_test.py
flixha: Thanks a lot for catching this. I got a bit confused because I had added two tests related to PR #439, of which one you already had merged into develop while the other one was part of the open PR. So I ended up thinking that my git history was messed up as it contained both tests for some reason; while it should have only contained the one which had already been merged. I reverted the last commit that removed `test_same_detections_individual_and_parallel` now, so the PR here should show only changes compared to the `develop`-branch now.

calum-chamberlain (approved):

Issue #455, opened 2021-04-15
Relative-magnitude: weight_by_correlation, min_cc & normalization are a bit confusing


flixha:
I was trying to calculate some relative magnitudes and was a bit confused by what `weight_by_correlation` and `min_cc` are supposed to do (or whether they are doing what they were intended to do).

# Main issue
Running `relative_magnitude` with `weight_by_correlation=True` returns delta-magnitudes that are each multiplied by the CC of the trace. So to calculate the actual magnitude, the user always has to do something like:
```python
from eqcorrscan.utils.mag_calc import relative_magnitude
delta_mags, ccs = relative_magnitudes(......, weight_by_correlation=True, return_correlations=True)
delta_mag_values = [_delta_mag[1] for _delta_mag in delta_mags.items()]
cc_values =  [cc[1] for cc in ccs.items()]
average_mag = original_mag + np.sum(delta_mag_values) / np.sum(cc_values)
```
Now I know that `family.relative_magnitude` is not working anyways right now, but looking into the code the normalization for the CC-values did not happen there as far as I can see. So it seems to me `relative_magnitude` wasn't intended to work like shown in the snippet above. Is that true?

# Other behavior I expected to work differently:
- By returning delta-mags multiplied with CC, if the user wants to save `StationMagnitude` for each observation, they first have to convert with `station_mag = original_mag + delta_mag / cc`
- When `weight_by_correlation=False`, then `min_cc` has no effect, i.e., delta-mags are returned for all traces, even those  that don't correlate well

# Code + Example
[Here is the code on how the CC-weighting works right now. ](https://github.com/eqcorrscan/EQcorrscan/blob/c3c2d44edbe0d5dcfa26a24da3aa742e07a03f7c/eqcorrscan/utils/mag_calc.py#L617-L636)

Here a minimal example based on the EQcorrscan-tests (this may not show the effect strongly, because the magnitude difference between event1 and event2 is rather small):
<details>
  <summary>(Click to expand minimal example!)</summary>
  
```python
import numpy as np
from obspy.clients.fdsn import Client
from eqcorrscan.utils.mag_calc import relative_magnitude

client = Client("GEONET")
event1 = client.get_events(eventid="2016p912302")[0]
event2 = client.get_events(eventid="3470170")[0]
shared_chans = {p1.waveform_id.get_seed_string()
                for p1 in event1.picks}.intersection(
    {p2.waveform_id.get_seed_string() for p2 in event2.picks})
[...]

PR #453, opened 2021-04-13
Test wildcards and ensure they work!

calum-chamberlain:
### What does this PR do?

Makes wildcard reading work for Party's

### Why was it initiated?  Any relevant Issues?

#387 - When using wildcards detections from multiple files, but the same family were not incorporated into the family (`=` rather than `+=`). This PR fixes this and adds a test.  @imankahbasi it would be great if you have time in the next day or so to try this out?

### PR Checklist
- [x] `develop` base branch selected?
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
~~- [ ] Any new or changed features have are fully documented.~~
- [x] Significant changes have been added to `CHANGES.md`.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.~~

PR #424, opened 2020-08-13
Patches for catalog-to-dd

calum-chamberlain:
## What does this PR do?

1. Check only the first char of the phase hint
2. Warn about different lengths being found, which occurs due to insufficient data passed
3. Skip correlations when no matching data are found.

### Why was it initiated?  Any relevant Issues?

@emilyws1 ran into more issues running catalog to dd when data were not trimmed correctly and phases had "odd" phase hints.

### PR Checklist
- [x] `develop` base branch selected?
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGES.md`.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.~~

Issue #365, opened 2019-12-14
Animation and plot of location


imankahbasi:
**Is your feature request related to a problem? Please describe.**
No. After detecting and locating new earthquakes It will be good to plot them to  observe and explore it to find some pattern if there is any.
Some functions like `threeD_seismplot` are good to make visual exploration.
But It is not enough.

**Describe the solution you'd like**
I think we need some more functions to plot and see the results of  the locations.
Something like 2D-plot of latitude and longitude with color mapping related to depth
or make 2D-animations to plot epicenters related to origin times and with color mapping related to origin time or depth.
I wrote some codes to provide these features.
I can transform these codes to functions and add them to this module if you would like.
Would you like to accept these?

**Additional context**
Color mapping respected to depth and longitude and latitude:
![depth](https://user-images.githubusercontent.com/47061763/70854214-05587180-1ece-11ea-8b4f-5975b1162e47.png)

Color mapping respected to origin time:
![time](https://user-images.githubusercontent.com/47061763/70854239-39cc2d80-1ece-11ea-8fd3-d4e31b234daf.png)

Animations of above plots:
https://www.dropbox.com/sh/tvqli5kqvez0fql/AAB8PppHUHYO0E8lFU8O-Mjna?dl=0

imankahbasi:
of course I should work on the time-dependent plot more.
Currently, this plot depicts according to the  occurring sequence and intervals are considered equal.
It is better to depict according to origin time.

calum-chamberlain:
Have you looked at obspy catalog plotting? It sounds like most of what you want is there.

CJ Chamberlain, out of office

________________________________

imankahbasi:
Yes, I saw that.
But it's not the same as what I want.

In obspy catalog plotting we can't change marker and marker size, I think it's good for plotting a few number of events but it's not good in case of too many events. In obspy the events overlap and we can't find density of them.

Obspy doesn't make cross section of events. It's very useful to make inspection and interpretation.

Obspy can't make animation of occurring events. I think it's very interesting and useful specially in case of presentation and maybe in interpretation.

Obspy uses catalog, but output of some software like Hypoellipse, perform the text file catalog. then it's good to have some functions that take only coordination of  points with optional time.

Obspy depends on the Basemap, but mine simply uses matplotlib.

calum-chamberlain:
Fair enough.  I do think that if you are going to do this then it would be worth using some projection to make sense of lat/long.

[...]

PR #457, opened 2021-04-19
Merge for 0.4.3 release

calum-chamberlain:
### What does this PR do?

Merge all the work for 0.4.3 release.

Multiple changes and bug-fixes for this release:
* core.match_filter
 - match_filter:
   - Provide option of exporting the cross-correlation sums for additional later analysis.
* core.match_filter.party.write
  - BUG-FIX: When `format='tar'` is selected, added a check for .tgz-file suffix before checking the filename against an existing file. Previously, when a filename without '.tgz'-suffix was supplied, then the file was  overwritten against the function's intention.
  - Add option `overwrite=True` to allow overwriting of existing files.
* core.match_filter.party.read
  - BUG-FIX: Ensure wildcard reading works as expected: #453
* core.match_filter.party.rethreshold:
  - added option to rethreshold based on absolute values to keep relevant detections with large negative detect_val.
* core.lag_calc:
  - Added option to set minimum CC threshold individually for detections based on: min(detect_val / n_chans * min_cc_from_mean_cc_factor, min_cc).
  - Added the ability of saving correlation data of the lag_calc.
* utils.mag_calc.calc_b_value:
  - Added useful information to doc-string regarding method and meaning of residuals
  - Changed the number of magnitudes used to an int (from a string!?)
* utils.mag_calc.relative_magnitude:
  - Refactor so that `min_cc` is used regardless of whether `weight_by_correlation` is set. See issue #455.
* utils.archive_read
  - Add support for wildcard-comparisons in the list of requested stations and channels.
  - New option `arctype='SDS'` to read from a SeisComp Data Structure (SDS). This option is also available in `utils.clustering.extract_detections` and in `utils.archive_read._check_available_data`.
* utils.catalog_to_dd
  - Bug-fixes in #424:
    - only P and S phases are used now (previously spurious amplitude picks were included in correlations);
    - Checks for length are done prior to correlations and more helpful error outputs are provided.
    - Progress is not reported within dt.cc computation
  - `write_station` now supports writing elevations: #424.
* utils.clustering
  - For `cluster`, `distance_matrix` and `cross_chan_correlation`, implemented full support for `shift_len != 0`. The latter two functions now return, in addition to the distance-matrix, a shift-matrix (both functions) and a shift-dictionary (for `distance_matrix`). New option for shifting streams as a whole or letting traces shift individually (`allow_individual_trace_shifts=True`).
* utils.plotting
  - Function added (twoD_seismplot) for plotting seismicity (#365).


calum-chamberlain:
All fails are NCEDC related. Merging.

Issue #445, opened 2021-03-15
What fundamental restrictions of templates, detection & lag-calc do we want to solve in the future?


flixha:
I'm opening this issue with the intent to collect information and discuss a suitable implementation strategy. So this is very much Work in Progress and I would appreciate any input to add to the current restrictions, related features wishes, and proposed implementation strategies.

**Current restrictions**
The current implementation of templates, detection, cross-correlation, and lag-calc has some restrictions including:

1. all template channels need the same length
2. P-picks are expected on Z-channel and S-picks on horizontal channels (some methods care only a bit about this, while for other methods it's fundamental
3. only one P- and one S-phase can be picked right now
4. all channels are weighted equally in cross-correlation 
5. computing `relative_magnitudes` doesn't work right now because templates don't contain enough information to compute pre-event noise (see #385)

**Is your feature request related to a problem? Please describe.**
It would be nice if we could support the following functionality in the future:

1. define templates with channels of "arbitrary" length
2. allow templates where P-picks and S-picks can appear on any channel (especially important when using single-component data)
3. support handling of additional phases like Pg, Pn etc. in template-creation, correlation, and lag-calc picking. (may want support for any kind of phase, e.g., teleseismic?)
4. allow arbitrary weighting of the template channels, and provide some standard ways to weight (e.g., equal weights; weighted by SNR)
5. store metadata like SNR in the template class

**Describe the solution you'd like**
For now, I'm opening this issue to collect a list of challenges, missing features, and requirements that would all be related, and hopefully addressed, in an update to the template class. I suspect that templates will need to contain some more metadata for us to allow the features described above. I think it would be desirable if we can get away with only one major upgrade to the template class so that we break compatibility with old templates only during one version update.

**Describe alternatives you've considered**
None yet - but this issue will likely evolve and I'm planning to fill these in here!

calum-chamberlain:
One thing that I would like to be able to do is have a template with all the information in it, but only use certain channels for detection, then all for subsequent lag-calc. I want to do this to be able to reduce computational requirements for long-running detection runs, but then be able to maintain the meta-data for picking at the end.

flixha:
[...]

PR #462, opened 2021-05-09
Parallel reading, chunked writing and catalog keying

calum-chamberlain:
### What does this PR do?

Implements a few speed-ups for `Party` IO, and ~~will~~ may include shifting to sparse event style objects. The main IO patches should be non-breaking and are:
1. Write chunked catalog files to the party archive, with a user-definable maximum number of events per file.
2. Read detection catalog files in parallel (optional), requires **1**
3. Move away from searching exhaustively (and stupidly) through the catalog of detection events for a matching id - just use a dictionary instead.

#445 mentions some of these points. Mostly speed for large catalogs was a real issue, and the catalog reading wasn't the main slow point! Scanning through the whole catalog to assign events to detections was dumb (my bad), and dictionary lookups speed this up a lot.

Writing out large catalogs can be really expensive for memory - it **SUCKS** getting to the end of a big matched-filter run, only to run out of memory when writing out the results... Limiting the number of events written to a single catalog file (hidden from the user in the tar archive) reduces the risk here - the main memory cost is holding both the catalog and the serialized catalog in memory at the same time - only serializing a chunk of the catalog at a time reduces the memory cost, without an obvious major slow-down. In theory this could be parallelized.

### PR Checklist
~~- [ ] `develop` base branch selected?~~ No, this is definitely a feature based on master.
- [ ] All tests still pass.


flixha (commented):
This looks pretty good to me!
I added some tiny comments for some parameters. I tried to test this a bit, and even though I don't have a good larger dataset to test this on right now, I was able to measure some speedups.

One issue I had: The parallel-reading works well for me in an isolated script, but running it as part of one of my detection routines, I ran into a multiprocessing-deadlock in using `multiprocessing.Pool`when reading a tribe. I got these deadlocks quite regularly on my system in my own functions until I started using spawning instead of forking the Process-pools (`multiprocessing.get_context("spawn").Pool()`). The deadlock here did not occur for me when spawning processes or using a `ThreadPool` (but these can be slower depending on context). From one quick test and some of my experience I would read the quakeml-files with a spawned pool, and the mseed-files with a `ThreadPool`***. 

Unfortunately, it's not so easy to provide a minimal working example for these deadlocks... 

***: If there's ever the possibility of corrupt mseed-files being read, then the ThreadPool would cause python to stop with an uncatcheable Segfault, while the process-pools can handle the error.

calum-chamberlain:
Thanks for those comments @flixha I agree and will get onto them! I'm not fond of setting the process spawning method, even in a context manager, mostly to ensure that the user *can* (but probably won't) set the spawn method.

Quick question, what OS and Python version are you running to get those deadlocks? I pretty much just develop on Linux and haven't run into deadlocks, but I know different OS's support multiprocessing in different ways. Would you also be able to share a script that deadlocks?

Finally, I may switch to using `concurrent.futures` to more easily allow `ProcessPoolExecutor` and `ThreadPoolExecutor` options for the reading of QuakeML and MSEED respectively.

eqcorrscan/core/match_filter/tribe.py
```diff
@@ -353,25 +370,68 @@ def read(self, filename):
...
-            tribe_cat = Catalog()
-        previous_template_names = [t.name for t in self.templates]
+            # Need to be able to link stream to template name
+            filenames = [t_files_dict.get(t_name)
+                         for t_name in template_names
+                         if t_name in t_files_dict.keys()]
+            # This should never fail due to above assertion...
+            assert len(filenames) == len(template_names), \
+                "Missing waveform files for templates"
+            with multiprocessing.Pool(processes=max_processes) as pool:
```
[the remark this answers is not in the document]
calum-chamberlain: Yup, this should be IO limited, and would be safer to use Threading, good point.

eqcorrscan/core/match_filter/party.py
```diff
@@ -803,17 +827,32 @@ def read(self, filename=None, read_detection_catalog=True,
...
-                    Logger.error(e)
-                    pass
+            Logger.info("Reading tribe")
+            tribe._read_from_folder(
+                dirname=party_dir, max_processes=max_processes)
+            det_cat_files = glob.glob(os.path.join(party_dir, "catalog.*"))
+            Logger.info("Reading detection catalog")
+            if len(det_cat_files) != 0 and read_detection_catalog:
+                if max_processes > 1:
+                    with multiprocessing.Pool(processes=max_processes) as pool:
```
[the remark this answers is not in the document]
calum-chamberlain: This definitely isn't IO limited, the main cost is CPU-bound deserialization of the xml structure in memory to the `Event` structure.

eqcorrscan/core/match_filter/party.py
```diff
@@ -754,7 +771,7 @@ def write(self, filename, format='tar', write_detection_catalog=True,
         return self
 
     def read(self, filename=None, read_detection_catalog=True,
-             estimate_origin=True):
+             estimate_origin=True, max_processes=1):
```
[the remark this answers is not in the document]
calum-chamberlain: I have changed to `cores`, but I haven't included the `parallel` argument, I regret that argument in general and revert to serial for `cores=1`.

calum-chamberlain:
An additional speed-up for party reading could be to read the parts of the party (Tribe, detection catalogs, detection stats) in separate processes, before finally reconstructing the party at the end.

I should provide some benchmarks for this at some point to actually demonstrate any speed-ups. One of the main speed-ups was actually the change from the silly lookup of template streams by an exhaustive search, to using a dictionary, which doesn't need any parallelism.

Initial testing showed that reading a Party of c. 300k detections and 2k templates went from taking > 9 hours to < 30 minutes. Still not *fast*, but at least more manageable...

eqcorrscan/core/match_filter/tribe.py
```diff
@@ -353,25 +371,77 @@ def read(self, filename):
...
-        if len(tribe_cat_file) != 0:
-            tribe_cat = read_events(tribe_cat_file[0])
+        t_files_dict = {os.path.splitext(os.path.basename(t))[0]: t
+                        for t in t_files}
+        if len(t_files) > len(t_files_dict):
+            Logger.warning("Duplicate template files found")
+        tribe_cat_files = glob.glob(os.path.join(dirname, "tribe_cat.*"))
+        tribe_cat = Catalog()
+        Logger.info("Reading tribe catalog")
+        if cores == 1 or len(tribe_cat_files) == 1:
```
flixha, line 400: I think here it would be useful to include a 
`cores = min(cores, len(tribe_cat_files))`
That would avoid the Pool to start more processes than necessary.

calum-chamberlain: See line 404, just below, before the pool is started.

calum-chamberlain: Bu you are right, I could just move that higher up and just check `if cores==1`.

flixha:
> Thanks for those comments @flixha I agree and will get onto them! I'm not fond of setting the process spawning method, even in a context manager, mostly to ensure that the user _can_ (but probably won't) set the spawn method.
> 
> Quick question, what OS and Python version are you running to get those deadlocks? I pretty much just develop on Linux and haven't run into deadlocks, but I know different OS's support multiprocessing in different ways. Would you also be able to share a script that deadlocks?
> 
> Finally, I may switch to using `concurrent.futures` to more easily allow `ProcessPoolExecutor` and `ThreadPoolExecutor` options for the reading of QuakeML and MSEED respectively.

I'm on python 3.8.8, and as OS I'm using RHEL 8.3 and RHEL 7.9 depending on the machine. I have seen many deadlocks especially when reading files in parallel on these. I've spent too much time trying to debug that without really understanding why it's happening. From what I read, python on Windows and Mac uses the spawn-method by default, and only on Linux uses forking by default, which is less robust in terms of package imports in the workers.

I wasn't fully aware of the context managing abilities of multiprocessing. I tried to use `multiprocessing.set_start_method('spawn')` at the start of the script below, but that will unfortunately result in a `RuntimeError: context has already been set`.


Example that will deadlock on the last line for me using the Linux machines; while it works on Mac 10.13.6:
```python
from obspy import UTCDateTime
from obspy.core.event import Pick
from eqcorrscan.utils import catalog_utils

client = Client('GEONET')
t1 = UTCDateTime(2016, 9, 4)
t2 = t1 + 86400
catalog = client.get_events(
    starttime=t1, endtime=t2, minmagnitude=4, minlatitude=-49,
    maxlatitude=-35, minlongitude=175.0, maxlongitude=-175.0)
catalog = catalog_utils.filter_picks(
    catalog, channels=['EHZ'], top_n_picks=5)
for event in catalog:
    extra_pick = Pick()
    extra_pick.phase_hint = 'S'
    extra_pick.time = event.picks[0].time + 10
    extra_pick.waveform_id = event.picks[0].waveform_id
    event.picks.append(extra_pick)
tribe = Tribe()
tribe.construct(
    method='from_client', catalog=catalog, client_id='GEONET',
    lowcut=2.0, highcut=9.0, samp_rate=50.0, filt_order=4,
    length=3.0, prepick=0.15, swin='all', process_len=3600)

tribe.write('test.tgz', max_events_per_file=4)

# Deadlocks here:
Tribe().read('test.tgz', cores=4)
```

calum-chamberlain:
Are you running that as a script or from an iPython session? In general when running scripts that use multiprocessing libraries you should always encapsulate your code in an `if __name__ == "__main__":` block. Check the multiprocessong docs [here](https://docs.python.org/3/library/multiprocessing.html#multiprocessing-programming) - the section on "Safe importing of main module" is what you need.

eqcorrscan/core/match_filter/tribe.py
flixha: Oops, sorry should have looked a bit further! I observed that it was using more processes than useful, but now realized that the branch wasn't correctly updated locally.

flixha:
The `if __name__ == "__main__":` did solve it indeed here!

calum-chamberlain:
I just ran that script without issue on my Ubuntu machine using Python 3.8.8 from conda... Curious that you run into locks. I have seen similar issues on Windows - there should probably be a note somewhere prominent in the docs about encapsulating everything in an `if __name__ == "__main__":` block. 

If you have the time/chance, if you run into these locks again without the block can you try that and report if it does/does not work?

Shameful admittance on my part, but I either run everything in iPython (for writing), Jupyter (for sharing with people), or write everything in functions by default because I tend to reuse a lot of code. I don't think I ever run EQcorrscan components in a script like this, hence not making any note of it!

flixha:
I get the deadlock both in ipython (in a window in VScode) and when running with python from the terminal. I have not generally had the deadlock-problem with EQcorrscan's functions (if I do run into that I can report it of course), but I have experienced it in several of my functions that wrap around EQcorrscan-functions (e.g., creating templates in parallel). In these cases, the deadlock still occurs even though all parallelization happens in the functions.

For the test above, if I put that into a function, it will run even without the `if __name__ == "__main__":`in front of the function call.

calum-chamberlain:
Interesting - I also don't use the iPython within vscode, or pycharm, I seem to get odd hangs in then in general, but if you are getting the same locks from the terminal then I don't know!

RE making templates in parallel - that may have something to do with pre-processing trying to process the streams in parallel and nested parallelism not working well. At some point I will redesign pre-processing to move away from multiprocessing - hopefully though some obspy speed-ups I'm playing with.
If you can, for parallel template creation, or anything that doesn't require any kind of memory sharing (e.g. looping through days of detections), I recommend not using Python multiprocessing to do this. I usually write a script with arguments of start and end and write a little bash script that calls the Python script multiple times and puts them into the background.

flixha:
Thanks for your thoughts on this!
When doing the template-creation in parallel, I use `parallel=False, cores=1` for the `pre_processing`. In total that makes it fast because most time in my scripts is spent on organizing picks and traces.

For the template-creation parallelization, I now use joblib `Parallel(n_jobs=cores)delayed(function)(parameters)`, which defaults to a robust pool-creation in loky through fork + exec instead of a simple fork or spawn in multiprocessing. That's more stable in that situation, while it comes with some other inconveniences, e.g., in  handling errors in the workers.

flixha:
Hi Calum,
I noticed an issue with the Resource-ID links when reading Quakeml files in parallel in subprocesses, see https://github.com/obspy/obspy/issues/2842 .
I haven't seen the problem in this implementation here yet, or it may not be particularly relevant, but thought it's worth mentioning in case it has a "hidden" effect.

eqcorrscan/core/match_filter/party.py
```diff
@@ -559,7 +561,16 @@ def decluster(self, trig_int, timing='detect', metric='avg_cor',
...
+
+        if min_chans > 0:
+            self.min_chans(min_chans - 1)
+            # min_chans uses explicit >, this uses >=
+
         all_detections = [d for fam in self.families for d in fam.detections]
+
+        if len(all_detections) == 0:
+            return Party()
+
```
calum-chamberlain, line 573: TODO: Check what these lines are about - they don't seem related to the topic of this PR?

PR #374, opened 2020-03-06
Patch catalog-to-dd

calum-chamberlain:
### What does this PR do?

Catalog to dd write_correlations uses parallel processing by default to process streams - multiprocessing on large machines seems to be running into memory errors (at SGEES).  This PR turns off default parallel processing, and shifts parallelism from Python to C for correlations.

Finn has been running into some other bugs with catalog-to-dd that we are patching at the moment, so this PR will stay open until those patches are included and tested.

### PR Checklist
- [x] `develop` base branch selected?
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are be covered via new tests.
- [ ] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGES.md`.
- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.

PR #473, opened 2021-09-22
catalog_to_dd.write_correlations: syntax fix for parallel process function call

flixha:
### What does this PR do?
- Adds a test for parallel execution of `_meta_filter_stream` in `eqcorrscan.utils.catalog_to_dd.write_correlations`
- Fixes how `apply_async` is called for `_meta_filter_stream`, which previously caused an error ( see below).
- Speeds up `write_correlations` by running `_prepare_stream` in parallel

Using `write_correlations` with `parallel_process=True, max_workers=2, lowcut=2.0` caused the following error for me (python 3.8.8, RHEL 8.2) on line 650: https://github.com/eqcorrscan/EQcorrscan/blob/90ee2e36cb8fdf2382acd997af2a2bd34da1ea07/eqcorrscan/utils/catalog_to_dd.py#L644-L650
```python
TypeError: _meta_filter_stream() got multiple values for argument 'stream_dict'
```
This can bee tested with this [test function ](https://github.com/flixha/EQcorrscan/blob/f1e467a145b72af82603ecc8fac80e4f81a841a3/eqcorrscan/tests/catalog_to_dd_test.py#L288-L300).

The error probably has to do with how and in which order `partial` and `apply_async` can accept arguments and keywords - I tried to fix it but didn't manage to find out how it could actually work. It works with the proposed change without using `partial()`.

In addition to the error, I noticed that three tasks that were not yet parallelized took a lot of time:
- the stream-filtering with `_meta_filter_stream` in `write_correlations`
- the stream-preparation with  `_prepare_stream` in `_compute_dt_correlations` before correlation
- the differntial-time calculation in `compute_differential_times` when using `write_catalog`
- Hence I added some code to allow `_prepare_stream` to be run in parallel. 

### PR Checklist
- [X] All tests still pass. (Plot tests fail right now locally due to HTTP error during data retrieval)
- ~[ ] Any new features or fixed regressions are be covered via new tests.`~
- ~[ ] Any new or changed features have are fully documented.~
- ~[ ] First time contributors have added your name to `CONTRIBUTORS.md`.~


flixha:
Around the time of when I opened this pull request, I started getting some strange errors for work on my fork of EQcorrscan ("remote rejected" , "remote: fatal: did not receive expected object", "corrupt remote". It may be safest not to download or merge this branch before I have figured out how to solve the issues with my fork...

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -184,7 +183,7 @@ def _prepare_stream(stream, event, extract_len, pre_pick, seed_pick_ids=None):
         elif len(pick) == 0:
             continue
         pick = pick[0]
-        tr = stream.select(id=seed_pick_id.seed_id).merge()
+        tr = stream.select(id=seed_pick_id.seed_id)
```
calum-chamberlain, line 186: Was there a reason for removing the merge?

calum-chamberlain (commented):
This looks good to me @flixha - just that one question about the `merge`, which I'm sure you had good reason for, but it just wasn't obvious to me!

eqcorrscan/utils/catalog_to_dd.py
flixha: I was running into some issues here, but I later reintroduced the merge with the newest commit, so please disregard this change!

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -11,7 +11,8 @@
 import numpy as np
 import logging
 from collections import namedtuple, defaultdict, Counter
-from functools import partial
+from obspy.core import stream
+from orderedset import OrderedSet
```
calum-chamberlain, line 15: This seems to be causing the tests to fail - `orderedset` doesn't exist - is this another dependency that needs to be added? Is there a way to not add this dependency and just run `list(set(thing))` to ensure order is retained?

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -135,10 +140,10 @@ def _generate_event_id_mapper(catalog, event_id_mapper=None):
     except ValueError:
         largest_event_id = 0
     for event in catalog:
-        if event.resource_id.id not in event_id_mapper.keys():
+        if str(event.resource_id) not in event_id_mapper.keys():
             event_id = largest_event_id + 1
             largest_event_id = event_id
-            event_id_mapper.update({event.resource_id.id: event_id})
+            event_id_mapper.update({str(event.resource_id): event_id})
```
calum-chamberlain, line 146: Was there a reason to change this? The `.__str__` method on `ResourceIdentifier` returns the `.id` property anyway (see [here](https://github.com/obspy/obspy/blob/master/obspy/core/event/resourceid.py#L672)) so it shouldn't make a difference which is called?

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -21,6 +22,7 @@
 from eqcorrscan.utils.clustering import dist_mat_km
 from eqcorrscan.core.lag_calc import _concatenate_and_correlate, _xcorr_interp
 from eqcorrscan.utils.correlate import pool_boy
+from wcmatch.glob import C
```
calum-chamberlain, line 25:
```suggestion

```

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -235,7 +242,11 @@ def _compute_dt_correlations(catalog, master, min_link, event_id_mapper,
     Logger.debug(f"Using channels: {master_seed_ids}")
     # Dictionary of travel-times for master keyed by {station}_{phase_hint}
     master_tts = dict()
-    master_origin_time = (master.preferred_origin() or master.origins[0]).time
+    try:
+        master_origin_time = (
+            master.preferred_origin() or master.origins[0]).time
+    except AttributeError:  # In case it's a SparseEvent
```
calum-chamberlain, line 248: Maybe `SparseEvent` should have a `preferred_origin` method that returns a `SparseOrigin`.

eqcorrscan/utils/catalog_to_dd.py
```diff
@@ -382,29 +409,67 @@ def _make_event_pair(sparse_event, master, event_id_mapper, min_link):
...
+def _prep_horiz_picks(catalog, stream_dict, event_id_mapper):
+    """
+    Fill in horizontal picks for the alternate horizontal channel for events in
+    catalog.
+    """
+    # keep user input safe
+    catalog = catalog.copy()
+    for event in catalog:
+        event_S_picks = [
+            pick for pick in event.picks if pick.phase_hint.startswith('S') and
```
calum-chamberlain, line 439:
```suggestion
            pick for pick in event.picks if pick.phase_hint.upper().startswith('S') and
```

calum-chamberlain:
You are right that parallel processing across traces for individual events will not load efficiently - I have generally avoided parallelism in this processing because I haven't got it to scale well (either on a per event basis, which can introduce really painful memory issues, or on a per-trace basis). I added options to turn off parallel processing in #374 because I was running into memory issues - as long as there is a way to not spawn pools then this should be fine though. I was generally trying to avoid Python parallel processing here - with 30k events I was unable to do anything with past parallel efforts.  What size dataset are you working with?

RE the changed station info - not me! But that is always a possibility with GeoNet changing their metadata. We will just have to update the test to cope with this new information.

eqcorrscan/utils/catalog_to_dd.py
flixha: By using `str()`, the function can take as an argument both the `SparseEvent` and the normal `Event`. With `event.resource_id.id`, it is only compatible with `obspy.core.event.Event`.

calum-chamberlain: Nice!

eqcorrscan/utils/catalog_to_dd.py
flixha: ups.. thanks for catching!

eqcorrscan/utils/catalog_to_dd.py
flixha: Right, I noted that when I first implemented it and indeed adding another dependency is not optimal. Found it a bit difficult to get my head around a good solution without the `OrderedSets` that wasn't several lines long, but now I realized it works with just one extra line of code.

eqcorrscan/utils/catalog_to_dd.py
flixha: The thing is that we then have to add the `preferred_origin_id`-property and the function itself which makes the event a bit less sparse... not a problem if you prefer it, but I understood that the `SparseEvent` should maybe be as sparse as possible?

eqcorrscan/utils/catalog_to_dd.py
flixha: fixed.

flixha:
thanks for the review and the comments! 
You're right that the scaling isn't great. I was using a dataset of ~870 events with 180 traces each, and was able to load at most ~15-20 cores out of 60, with each process using 10 GB memory (total ~150-200 GB) while 512 GB were available. Parallelization across events did make it ~2.5 times quicker for this very dataset. It's something, but clearly parallelization inside Python may not be optimal.

Re station info: Do you know who maintains the station metadata at GEONET? It's a small mistake that can easily happen, but it should best be fixed at the source so people can continue using data from CTZ (if they're running Seiscomp, then with the metadata for pre-2021 removed no data can be downloaded for the station any more). Changing the test is possible, but it becomes a bit hacky if we manually remove the comparison for that station (or change to an event that doesn't include CTZ).

eqcorrscan/utils/catalog_to_dd.py
calum-chamberlain: Yeah, I think you are right - sparsity is good.

calum-chamberlain:
Can you confirm that you are still able to run without event parallelization and with similar memory consumption? If so, then I am keen to keep your changes in :) - we might want to add a note that if the user runs into memory limitations for large datasets then they will need to turn to trace parallelism, and a note on how to do that.

I do know who maintains the metadata at GeoNet - and they have an [issues page on github](https://github.com/GeoNet/help/issues/92) and I have opened a new issue - thanks for spotting this @flixha - I thought you meant that the co-ordinates had been changed rather than it not being there anymore! The team at GeoNet are usually very responsive, and we might wait for them to update their services before merging this to get all the green lights.

flixha:
Yes, if I use only trace-parallelization (`max_trace_workers=40, max_workers=1`), then it uses ~7 GB memory in total for the dataset (865 events), while if I use event-parallelization it uses ~5 GB per worker (so 100 GB for 20 workers; the 10 GB per worker that I wrote earlier must be 50 % shared memory and 50 % "new" memory per worker). I included a test [here](https://github.com/eqcorrscan/EQcorrscan/blob/23cf7c22154b2bffe0113bdc930f5939d73f4d79/eqcorrscan/tests/catalog_to_dd_test.py#L309-L322) to check that trace-parallelization works.

Regarding the metadata: Great, thanks for opening the issue on the GeoNet issues page, and they're indeed so quick that it's fixed already!  I just triggered another test run that should then run without the error...

I put a [note here](https://github.com/eqcorrscan/EQcorrscan/blob/23cf7c22154b2bffe0113bdc930f5939d73f4d79/eqcorrscan/utils/catalog_to_dd.py#L514-L535) to inform users about the parallelization options and memory consumption - let me know if you think this should go somewhere else.

calum-chamberlain:
Golden, that looks good to me, thanks for checking! Are you happy for me to merge this now? The failing tests are unrelated network read errors.

flixha:
Yes, I'm happy to have this merged - thanks a lot for reviewing!

PR #491, opened 2022-03-25
Make CC-interpolation simpler & more robust

flixha:
### What does this PR do?
Replaces the CC-interpolation function that can be used for `lag_calc` and `catalog_to_dd` with a function that simply resamples the CC.

- CC-interpolation with `_xcorr_interp` "failed" (i.e., no improvement with interpolation) more often than maybe necessary. For the example data in the test, the new function fails to produce an improved CC with interpolation in only about 90 cases, while the previous one failed in about 150 cases. 
- `_xcorr_interp` had one specific weakness: it needs more than 2 samples with negative curvature to interpolate around the peak of the CC. If seismograms are filtered just below the Nyquist frequency (e.g., sampling rate 20 Hz and low-pass filter at 9.9 Hz) and contain plenty of high-frequency content just below the Nyquist frequency, then that criterion of more than 2 samples with negative curvature is often not fulfilled [(c.f. these lines)](https://github.com/eqcorrscan/EQcorrscan/blob/6286e4ff2dbb073eb98379ea69390ff3841b0b09/eqcorrscan/core/lag_calc.py#L74-L78) 
- On a dataset where the above was quite relevant, the new function improved the dt-accuracy for about 15 % more correlations than the old function (n=8*10^7)

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ## Current
 * core.match_filter.tribe
  - Detect now allows passing of pre-processed data
+* core.lag_calc._xcorr_interp
+ - -CC-interpolation replaced with resampling (more robust)
 * utils.correlate
  - Fast Matched Filter now supported natively for version >= 1.4.0
  - Only full correlation stacks are returned now (e.g. where fewer than than
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@
 * core.match_filter.tribe
  - Detect now allows passing of pre-processed data
 * core.lag_calc._xcorr_interp
- - -CC-interpolation replaced with resampling (more robust)
+ - CC-interpolation replaced with resampling (more robust), old method
+   deprecated. Use new method with use_new_resamp_method=True as **kwarg.
 * utils.correlate
  - Fast Matched Filter now supported natively for version >= 1.4.0
  - Only full correlation stacks are returned now (e.g. where fewer than than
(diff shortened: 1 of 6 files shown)


calum-chamberlain:
This looks nice! Although I think this should go through a deprecation cycle - e.g. with an option to use the resampling method rather than the current method. I think the current method is documented in a couple of spots linking to a paper so we would need to find and change those as well.

flixha:
@calum-chamberlain Of course, I'll fix this so that the old methods stays as the default with a deprecation message for a while, and users can choose this method. Should have mentioned: I saw that there was a [comment](https://github.com/eqcorrscan/EQcorrscan/blob/6286e4ff2dbb073eb98379ea69390ff3841b0b09/eqcorrscan/core/lag_calc.py#L67) in the function that you may have put in because you saw that it doesn't always work optimally - any more thoughts on that that we should consider for the new function?

flixha:
I made the new function optional with `use_new_resamp_method=True` which is passed as `**kwargs`. The warning message is only shown once so that the log isn't completely spammed with this message - because this function is called many times when interpolation is turned on.

flixha:
Tests now failing during installation, probably related to a breaking change in setuptools 61.0.0: https://github.com/pypa/setuptools/issues/3197 .. checking this now.

calum-chamberlain (approved):
Looks good to me - if we can get this in the next release then we can get the deprecation warning out of the way quickly.  If you are happy I will merge this.

flixha:
Sounds good, thanks for merging!

PR #493, opened 2022-04-22
WIP: Bug fix where lag_calc correlation window starts slightly too late

flixha:
### What does this PR do?
- Fixes a bug where streams start too late for `lag_calc` because it doesn't take `prepick` into account
- Fixes another small bug for option `min_cc_from_mean_cc_factor` introduced in #446 when the `detect_val` is negative

I noticed that when I generate templates with `prepick=0.5` and then run `lag_calc` with `shift_len=0.4` (i.e., `shift_len < prepick`), then `lag_calc` does not return any useful picks (see example below).
If `shift_len > prepick` then `lag_calc` returns perfect picks for a template that detects itself, but for picking new detection this bug may introduce a bias where picks are set on average too late because the correlation window for `lag_calc` is not centered around the detection time.

**Relevant issues for which it would be great to get some feedback before I complete this PR (tests, fixes, docs..):**
- to let `lag_calc` set up the correlation window correctly, it needs to know the `prepick`-value that is not part of the template matadata
- now I let `lag_calc` infer `prepick` by comparing the picks in the template event to the trace's starttime
- that works when the event is saved with the template. I noticed some synthetic tests did not have that set up (which can be fixed), but are there other cases that I should take into account here?
- are we happy with how this is handled 

Here a demonstration with the example from the tutorial (will not return any picks without bug fix):

```python
import logging
logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s\t%(name)s\t%(levelname)s\t%(message)s")

from obspy import UTCDateTime

client = Client('NCEDC')
catalog = client.get_events(eventid='72572665', includearrivals=True)
catalog[0].picks = catalog[0].picks[0:5]
tribe = Tribe().construct(
     method='from_client', catalog=catalog, client_id='NCEDC', lowcut=2.0,
     highcut=8.0, samp_rate=20.0, filt_order=4, length=6.0, prepick=0.5,
     swin='all', process_len=3600, all_horiz=True)

party, stream = tribe.client_detect(
     client=client, starttime=UTCDateTime(2016, 1, 2),
     endtime=UTCDateTime(2016, 1, 3), threshold=8, threshold_type='MAD',
     trig_int=6, plotvar=False, return_stream=True)

stream = stream.merge().sort(['station'])
repicked_catalog = party.lag_calc(stream, pre_processed=False,
                                  shift_len=0.4, min_cc=0.5)
```

### PR Checklist
- [X] All tests still pass.


flixha:
This PR is resolved much cleaner with #500, so will close it soon..

calum-chamberlain:
Ah darn, sorry I forgot about this one. Can you include the `min_cc_from_mean_factor` change in another PR?

flixha:
> Ah darn, sorry I forgot about this one. Can you include the `min_cc_from_mean_factor` change in another PR?
This is now in https://github.com/eqcorrscan/EQcorrscan/pull/504, so I'll close here.

Issue #497, opened 2022-06-20
Template write / read merges traces in rare case where traces from same channel align exactly


flixha:
**Describe the bug**
Consider a template for a 1-component sensor with only a vertical channel, with P- and S-pick on the same trace. If the traces happen to line up exactly, that is, the first sample of the S-trace follows the last sample of the P-trace, then the obspy / mseed library merges the waveforms for the P- and S-part into one trace. The result is a template where some traces suddenly have double the length, and once there is a mix of trace lengths, then of course EQcorrscan will report an error.

**To Reproduce**
see attached template object - trace NS.ARA1.00.BHZ is merged into one trace when reading, even though the template had two traces for NS.ARA1.00.BHZ before writing.
[Template_sample.gz](https://github.com/eqcorrscan/EQcorrscan/files/8939175/Template_sample.gz)  (.tgz file renamed to .gz due to file type restrictions on Github)
```python
from eqcorrscan.core.match_filter import Template, Tribe
tribe = Tribe().read('Template_sample.gz')
templ = tribe[0]
assert len(list(set([tr.stats.npts for tr in templ.st]))) == 1
```

**Expected behavior**
The trace should be properly split into the two parts for P and S.

**Desktop (please complete the following information):**
 - Operating System: RHEL 8.3
 - Python version: 3.9
 - EQcorrscan version: develop

**Additional context**
This problem is not directly caused by EQcorrscan, but rather by the mseed-reading routines. I'm trying to look for some lower-level options that stop how the data are merged into one trace in the miniseed routines, but I don't know the codes so well so this may take some time. I'm not sure yet whether this is an issue of reading or writing the miniseed file. If the mseed library doesn't allow the trace to be split in this case, then we may have to think of a workaround or some extra checks.

Even though I'm continuing to investigate, if you have any ideas, please let me know!

calum-chamberlain:
Nice spot! That's an ugly bug, thanks for finding it.  I think we should probably try to cope with it in here if we can, although I'm not sure how at the moment....

Sorry for the bug!

ylseanna:
+1

I annoyingly also encountered this problem, twice during a single day of data even. Any fixes on the horizon (or known work-arounds in the meantime)?

PR #498, opened 2022-06-20
WIP: Fix bug where perfectly aligning traces are merged into one trace by MSeed read routine

flixha:
### What does this PR do?
First attempt at fixing #497. This checks that all template traces have the same length. If it finds any traces that have double (or another multiple) of the "normal" template length, then it splits these traces into traces of the exact same length with the newly corrected start and endtimes. 

I will add testing / changes-log once I / we feel that this is the right way to avoid the bug.

The fix here keeps all the checks within EQcorrscan, while in theory I was hoping to find an option that avoids the behavior that leads to this bug in the mseed-routines. But I don't think that such an option exists there for exact matches; the only option I do know is that one case change a hard-coded option in the miniseed-library to reduce the tolerance until which traces are merged. But that only concerns the matching of time samples; not block numbers as far as I know.

As we don't (yet) save template length as a template / tribe metadata, the following edge cases would not be solved by this PR:
 - when all traces of a template are affected by the mseed-segment merging problem
 - maybe other sample 

### PR Checklist
- [ ] All tests still pass.
- ~[ ] First time contributors have added your name to `CONTRIBUTORS.md`.~


calum-chamberlain:
This is a nice idea to cope with this, but frustratingly the lack of checks for short template lengths in other places can lead to some short traces being included in templates. This means that taking the minimum of the trace lengths to be the correct length probably won't work as a general fix.

I wonder if it is time to add some more details to the template files that are written out, like the nominal length, or the start and end times of each channel in the template.

It would also be good to catch the issue of incorrect template lengths on creation...

flixha:
I see that; I didn't fully think through that part about templates that are too short because I had my own checks in place while I was creating the templates. So this solution here won't work on its own for now..

calum-chamberlain:
Sadly no - I think that generating templates not of the length requested is a bug that should be fixed as well, but that still won't help with older templates, or templates that someone might have constructed not using EQcorrscan (once the bug is fixed). In another issue (I forget which, sorry), allowing different lengths on different channels has been discussed as well. Maybe we should just commit to fixing this by adding more metadata and forcibly reconstructing the waveforms to meet that metadata?

PR #524, opened 2022-12-12
Speedup 01 group templates

flixha:
### What does this PR do?
 Gives 50x speed up for grouping templates.

Grouping many templates was slow because the function was using a double loop for template-parameter comparison (O^2). So grouping took ~20 s for ~2000 templates. Sped up single loop to <0.5 s.

The new, quick function now circumvents the repeated calls to `template.same_processing` that contribute to the slowdown. So if we change what `template.same_processing` does (i.e., which dict-entries it compares), then we'd also need to change it here. Left in the old function for testing for now.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/eqcorrscan/core/match_filter/template.py
+++ b/eqcorrscan/core/match_filter/template.py
@@ -714,25 +714,26 @@ def quick_group_templates(templates):
     :type templates: List of Tribe of Templates
     :return: List of Lists of Templates.
     """
-    # Get the template's processing parameters and a unique list of
-    # parameter-combinations
+    # Get the template's processing parameters
     processing_tuples = [
         tuple(
             [value for key, value in template.__dict__.items()
              if key not in ['name', 'st', 'prepick', 'event', 'template_info']]
             )
         for template in templates]
-    # Convert to numpy array to efficiently search for rows with same values
-    processing_array = np.array(processing_tuples)
+    # Get list of unique parameter-tuples. Sort it so that the order in which
+    # the groups are processed is consistent across different runs.
     uniq_processing_parameters = sorted(list(set(processing_tuples)))
     # sort templates into groups
     template_groups = []
     for parameter_combination in uniq_processing_parameters:
-        # find indices of rows with same parameters
-        template_indices_for_group = np.where(
-            (processing_array == np.array(parameter_combination)).all(axis=1))
+        # find indices of tuples in list with same parameters
+        template_indices_for_group = [
+            j for j, param_tuple in enumerate(processing_tuples)
+            if param_tuple == parameter_combination]
+
         new_group = list()
-        for template_index in template_indices_for_group[0]:
+        for template_index in template_indices_for_group: #[0]:
             # use indices to sort templates into groups
             new_group.append(templates[int(template_index)])
         template_groups.append(new_group)
--- a/eqcorrscan/core/match_filter/template.py
+++ b/eqcorrscan/core/match_filter/template.py
@@ -707,6 +707,38 @@ def group_templates(templates):
     return template_groups
 
 
+def quick_group_templates(templates):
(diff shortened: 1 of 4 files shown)


flixha:
NCEDC - FDSN not answering properly in tests..

eqcorrscan/core/match_filter/template.py
```diff
@@ -707,6 +707,39 @@ def group_templates(templates):
...
+    # sort templates into groups
+    template_groups = []
+    for parameter_combination in uniq_processing_parameters:
+        # find indices of tuples in list with same parameters
+        template_indices_for_group = [
+            j for j, param_tuple in enumerate(processing_tuples)
+            if param_tuple == parameter_combination]
+
+        new_group = list()
+        for template_index in template_indices_for_group: #[0]:
```
calum-chamberlain, line 736:
```suggestion
        for template_index in template_indices_for_group:
```

eqcorrscan/core/match_filter/tribe.py
```diff
@@ -24,7 +24,8 @@
 from obspy import Catalog, Stream, read, read_events
 from obspy.core.event import Comment, CreationInfo
 
-from eqcorrscan.core.match_filter.template import Template, group_templates
+from eqcorrscan.core.match_filter.template import (
+    Template, group_templates, quick_group_templates)
```
calum-chamberlain, line 28:
```suggestion
    Template, quick_group_templates)
```

calum-chamberlain (approved):
Looks great to me! thanks for this.

eqcorrscan/tests/match_filter_test.py
```diff
@@ -586,6 +586,19 @@ def test_tribe_copy(self):
             self.assertEqual(t, copy_t)
         self.assertEqual(tribe, copied)
 
+    def test_tribe_copy_quick(self):
+        from robustraqn.templates_creation import _quick_tribe_copy
```
calum-chamberlain, line 590:
```suggestion
        from eqcorrscan.core.match_filter.template import quick_group_templates
```

eqcorrscan/tests/match_filter_test.py
```diff
@@ -586,6 +586,19 @@ def test_tribe_copy(self):
             self.assertEqual(t, copy_t)
         self.assertEqual(tribe, copied)
 
+    def test_tribe_copy_quick(self):
```
calum-chamberlain, line 589: This looks like the wrong test for this PR.

calum-chamberlain:
For some reason I'm getting segfaults in this branch - develop doesn't seem to have this issue. @flixha are you also getting segfaults for this branch? Mine are coming from the NCEDC test Cases.

calum-chamberlain:
Looks good to me @flixha - point noted on the `template.same_processing`. How about we make a `template._processing_parameters` property and use that instead of the:
```python
if key in ["name", "st", "prepick", "event", "template_info"]:
    continue
else:
    actually compare
```
if logic in `template.same_processing`? Then your `processing_tuples` could just access those defined params. Any changes to the params compared in `template.same_processing` would then have to be made to the property and hence to your func?

eqcorrscan/tests/match_filter_test.py
flixha: Ups.. thanks for removing!

flixha:
> `template._processing_parameters` property and use that instead of the:

Very good suggestion. Now I defined the `_processing_parameters` as function / property for a Template, which helps to simplify the code at three places so that it's simpler to update any changes to what we consider processing parameters.

calum-chamberlain (approved):
Nice, that is what I was thinking, glad you agree. Looks like that should be simpler to maintain.

Issue #522, opened 2022-12-08
WIP: Speed up a few slowdowns when handling large datasets


flixha:
This is a summary thread for a few slowdowns that I noticed when handling large-ish datasets (e.g., 15000 templates x 500 channels x 1800 samples). I'm not calling them "bottlenecks" because it's rather the sum of things together that take extra time, none of these slowdowns changes EQcorrscan fundamentally.

I will create some PRs for each point that I have a suggested solution so that we can systematically merge, improve, or reject the suggested solutions. Will add the links to the PRs here, but I'll need to organize a bit for that..

Here are some slowdowns (tests with python 3.11, all in serial code):

- `tribe._group_templates`: **50x speed up**
  -  problem O^2 double loop takes ~20 s for some thousand templates. Sped up single loop to <0.5 s.
- `preprocessing._prep_data_for_correlation`: **3x speed up**
  - Problem: with heterogeneous templates (i.e., many templates with different station setups), filling the templates with NaN-channels takes a long time (many copy-operations in serial). (ca. 150 s -> 50 s for 1500 templates with up to 500 channels).
- `matched_filter.match_filter` with `copy_data=True`: **3x speedup for copy**
  - copying of streams (templates and continuous data) takes XYZ s because deepcopy of streams/traces is slow. Custom copy functions can speed it up by a factor of ~3x. Circa 18 s previously for 300 24h- traces, now ~6 s.
- `detection`, `lag_calc`,  `pre_processing`: **4x / 100x speedup for trace selection**
  - trace selection by trace id from a stream is still a slowdown (even after https://github.com/obspy/obspy/pull/2886). Can be speed up by x4 with simplified function and ~x100 with dict-lookup for streams with fixed order of traces.
- `core.match_filter.family._uniq`: **1.9x speedup**
  - retrieving the unique list of detections is quicker for many detections with `list(set)` (1.9x speedup for 43000 detections, fastest: 3.1 s), but 1.2x slower for small sets (e.g., 430 detections; 50 ms --> 27 ms).
- `core.match_filter.detect` - **1000x speed up for many calls to `family._uniq`**
  -  using `family._uniq` in a loop over all families is still rather slow with `_uniq`. Checking tuples of `(detection.id, detection.detect_time, detection.detect_val)` with `numpy.unique` and avoiding a loop is 1000x faster.  From 752 s to <1 s for 82000 detections.
- `_group_detect`: **30x speedup in handling detections**
[...]

PR #525, opened 2022-12-12
Speedup 02: 3x speed up for prep_data_for_correlation with custom copy and trace-selection

flixha:
### What does this PR do?
- 3x speed up for function `preprocessing._prep_data_for_correlation`
- provides new functions `_quick_copy_trace` and `_quick_stream_copy` that are ~3x quicker for copying simple traces by avoiding a call to `copy.deepcopy` where that is not necessary. So it helps with traces that mostly contain data, but will not speed up copying as much for traces with attached response and longer history.
 
When a user has a big set of heterogeneous templates (i.e., many templates with different station setups), filling the templates with NaN-channels takes a long time (many copy-operations in serial). This PR speeds that process up by a factor of 3 in my example, going from ca. 150 s to 50 s for 1500 templates with up to 500 channels.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,8 @@
 ## Current
+* utils.pre_processing:
+  - _prep_data_for_correlation: 3x speedup for filling NaN-traces in templates
+  - new functions _quick_copy_trace and _quick_stream_copy for 3x quicker
+    trace / stream copy.
 
 ## 0.4.4
 * core.match_filter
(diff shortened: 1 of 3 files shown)


calum-chamberlain:
This looks good, but it looks like it is missing the test from #524

calum-chamberlain (approved):

calum-chamberlain:
For some reason I'm getting segfaults in this branch - develop doesn't seem to have this issue. @flixha are you also getting segfaults for this branch? Mine are coming from the NCEDC test Cases.

flixha:
> For some reason I'm getting segfaults in this branch - develop doesn't seem to have this issue. @flixha are you also getting segfaults for this branch? Mine are coming from the NCEDC test Cases.
> 
> I expect that this is due to some unforeseen change in how data are now being passed to the C-funcs. Can you have a look please?

I just found that somehow one commit got lost along the way that had fixed an issue that turned up here again.. Basically it was a mistake on my side where I did not properly set the entries in the stats-dict for the NaN-channel. Let's see how the tests here are doing now; it works for me locally as it did at some point before.

calum-chamberlain (approved):

PR #526, opened 2022-12-12
Speedup 04: quick trace selection

flixha:
### What does this PR do?
-  4x / 100x speedup for trace selection in `detection`, `lag_calc`,  `pre_processing`

-  trace selection by trace id from a stream is still something that takes more time than it needs to when working with many thousand (100K?) traces, even after https://github.com/obspy/obspy/pull/2886.
- When a user doesn't need wildcard selection, than the simple selection function can be sped up by a factor of 4 compared to `obspy.stream.select` 
- For the time when the order of traces in a stream does not change and a lot of traces need to be selected from that stream, then it's possible to speed up the selection by a factor of ~100 with a dict-lookup - this is very relevant for the template preparation in `_prep_data_for_correlation`

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,7 @@
 ## Current
+* utils.pre_processing
+  - New function ``quick_trace_select` for a very efficient selection of trace
+    by seed ID without wildcards (4x speedup).
 
 ## 0.4.4
 * core.match_filter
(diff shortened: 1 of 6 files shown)


calum-chamberlain (approved):
Looks great to me and tests were passing. Nice speedup!

PR #527, opened 2022-12-12
Speedup 05: Retrieve unique detections in family and in `matched_filter`

flixha:
### What does this PR do?
- `core.match_filter.family._uniq`: **1.9x speedup**
  - retrieving the unique list of detections is quicker for many detections with `list(set)` (1.9x speedup for 43000 detections, fastest: 3.1 s), but 1.2x slower for small sets (e.g., 430 detections; 50 ms --> 27 ms).
- `core.match_filter.detect` - **1000x speed up for many calls to `family._uniq`**
  -  using `family._uniq` in a loop over all families is still rather slow with `_uniq`. Checking tuples of `(detection.id, detection.detect_time, detection.detect_val)` with `numpy.unique` and avoiding a loop is 1000x faster.  From 752 s to <1 s for 82000 detections.

Retrieving unique detections for `matched_filter`-run was getting slower than needed when there are a lot of detections.

### PR Checklist
- [ ] All tests still pass.
~- [] Any new or changed features have are fully documented.~

### DIFF
--- a/eqcorrscan/core/match_filter/family.py
+++ b/eqcorrscan/core/match_filter/family.py
@@ -297,7 +297,6 @@ class Family(object):
         """
         return len(self.detections)
 
-
     def _uniq(self):
         """
         Get list of unique detections.
--- a/eqcorrscan/core/match_filter/family.py
+++ b/eqcorrscan/core/match_filter/family.py
@@ -328,8 +328,6 @@ class Family(object):
         """
         _detections = []
         _detections = list(set(self.detections))
-        # [_detections.append(d) for d in self.detections
-        # if not _detections.count(d)]
         self.detections = _detections
         return self
 
--- a/eqcorrscan/core/match_filter/family.py
+++ b/eqcorrscan/core/match_filter/family.py
@@ -326,8 +326,8 @@ class Family(object):
         2
         """
         _detections = []
-        _detections = [_detections.append(d) for d in self.detections
-                       if not _detections.count(d)]
+        [_detections.append(d) for d in self.detections
+         if not _detections.count(d)]
         self.detections = _detections
         return self
 
--- a/eqcorrscan/core/match_filter/family.py
+++ b/eqcorrscan/core/match_filter/family.py
@@ -297,6 +297,7 @@ class Family(object):
         """
         return len(self.detections)
 
+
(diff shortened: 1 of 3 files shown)


eqcorrscan/core/match_filter/family.py
```diff
@@ -326,8 +326,7 @@ def _uniq(self):
         2
         """
         _detections = []
-        [_detections.append(d) for d in self.detections
-         if not _detections.count(d)]
+        _detections = list(set(self.detections))
```
calum-chamberlain, line 329: Making a `set` of detections relies on them being hashable, which they are not (or should not be - they can and do change - I did accidentally make them hashable at one point). Therefore I don't think this method should work. If it isn't returning the following error:
```python
TypeError: unhashable type: 'Detection'
```
then it should!

eqcorrscan/core/match_filter/tribe.py
```diff
@@ -606,7 +606,21 @@ def detect(self, stream, threshold, threshold_type, trig_int, plot=False,
...
+                        for det in family]
+                    # Retrieve the indices for the first occurrence of each
+                    # detection in the family (so only unique detections will
+                    # remain).
+                    uniq_det_tuples, uniq_det_indices = np.unique(
+                        det_tuples, return_index=True, axis=0)
+                    uniq_detections = []
+                    for uniq_det_index in uniq_det_indices:
+                        uniq_detections.append(family[uniq_det_index])
+                    family.detections = uniq_detections
```
calum-chamberlain, line 623: This should be fine (for what we want, which is no multiple detections at the same time for the same template and hence family, this is ideal) - and a lot faster! Nice!

calum-chamberlain (changes requested):
It looks like the changes to `_uniq` shouldn't work, and if they do then we should make sure they don't... Otherwise this is a really useful speedup!

eqcorrscan/core/match_filter/family.py
flixha: Thanks for pointing this out.. indeed I didn't get the error in any of my tests which is surprising considering that these errors came up in all tests on github. I'll revert this change here

calum-chamberlain (approved):
Golden, thanks!

PR #533, opened 2022-12-22
Update test_conda_env.yml

calum-chamberlain:
trying to fix the test env.


calum-chamberlain:
This took a lot of stuffing about to get osx to run again - the build env that conda-forge uses changed but it is now running in a similar env.

The failing slow tests relate to subspace issues which relate to changes to numpy handling of array setting with sequence in `core.subspace.multi`. This should be a simple fix for another PR.

PR #536, opened 2023-01-10
Speedup 10:  25 % quicker _get_array_dicts and 10 % quicker _prep_data_for_correlation

flixha:
### What does this PR do?
Implements:
- 25 % speedup for `utils.correlate._get_array_dicts` (from 200 s to 154 seconds total time spent in my test)
- 10 % speedup for `utils.pre_processing._prep_data_for_correlation` (from 702 s to 640 s total time spent)
- (for comparison to the numbers above: 494 s were spent for correlating data in parallel on the GPU)

Main speedups with:
- direct operations on `UTCDateTime.__dict__['_UTCDateTime__ns']` (5x quicker for one +/- operation)
- index-based access to trace in stream with `stream.traces[j]` instead of `stream[j]` (4.5x quicker)

For big datasets run on powerful hardware, the cumulative time spent in `utils.correlate._get_array_dicts` and `utils.pre_processing._prep_data_for_correlation` makes up a considerable part (>50 %) of a detection run. This PR tries to optimize the functions with some simple changes; I think further speed ups would need changes that are not so trivial...

This PR is part of the list of speedups in https://github.com/eqcorrscan/EQcorrscan/issues/522

### PR Checklist
- [] All tests still pass.

### DIFF
--- a/eqcorrscan/utils/pre_processing.py
+++ b/eqcorrscan/utils/pre_processing.py
@@ -1000,7 +1000,7 @@ def _prep_data_for_correlation(stream, templates, template_names=None,
             if len(template_channel) <= channel_index:
                 # out_template[channel_number].data = nan_channel  # quicker:
                 out_template.traces[channel_number].__dict__[
-                    'data'] = copy.deepcopy(nan_channel)
+                    'data'] = np.copy(nan_channel)
                 out_template.traces[channel_number].stats.__dict__[
                     'npts'] = template_length
                 out_template.traces[channel_number].stats.__dict__[
--- a/eqcorrscan/utils/pre_processing.py
+++ b/eqcorrscan/utils/pre_processing.py
@@ -1008,7 +1008,7 @@ def _prep_data_for_correlation(stream, templates, template_names=None,
                 out_template.traces[channel_number].stats.__dict__[
                     'endtime'] = UTCDateTime(ns=int(
                         round(template_starttime.ns
-                              + (template_length * samp_rate) * 1e9)))
+                              + (template_length * samp_rate) / 1e9)))
             else:
                 out_template.traces[channel_number] = template_channel.traces[
                     channel_index]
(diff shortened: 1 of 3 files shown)


flixha:
Sorry for the mess with 2 commits with the same message; I thought I had reset / removed the first one locally before pushing. As it's so small I think it's easier to leave it than trying to completely remove that commit from remote; anyways it's fixed by the last commit.

eqcorrscan/utils/pre_processing.py
```diff
@@ -999,18 +999,20 @@ def _prep_data_for_correlation(stream, templates, template_names=None,
...
-                out_template[channel_number].__dict__['data'] = copy.deepcopy(
-                    nan_channel)
-                out_template[channel_number].stats.__dict__['npts'] = \
-                    template_length
-                out_template[channel_number].stats.__dict__['starttime'] = \
-                    template_starttime
-                out_template[channel_number].stats.__dict__['endtime'] = \
-                    UTCDateTime(ns=int(
+                out_template.traces[channel_number].__dict__[
+                    'data'] = copy.deepcopy(nan_channel)
```
calum-chamberlain, line 1003:
```suggestion
                    'data'] = np.copy(nan_channel)
```
I wondered if there was a faster numpy way to copy arrays, and there is (although not faster). On my machine:

```python
import numpy as np
import copy

a = np.random.randn(8640000)
%timeit copy.deepcopy(a)
# 11 ms ± 263 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit np.copy(a)
# 10 ms ± 166 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
There is very little in it, but I can't see a good reason not to us `np.copy` (I just didn't know about it until now!).

calum-chamberlain (approved):
Tiny change that I will push through and merge if this goes green.

eqcorrscan/utils/pre_processing.py
flixha: This is a very good point, should definitely contribute to the overall speedup. Thanks for finding this and merging!

PR #541, opened 2023-03-19
[WIP] Group templates by seed id

calum-chamberlain:
### What does this PR do?

Groups templates by SEED ID when using grouping. This should reduce the number of nan-traces used and make the correlations and processing more efficient. On opening this PR, I have not yet tested the code, this serves as a suggestion/reminder of the logic I thought of this morning.

Mentioned in #522 @flixha requested something like this.

### PR Checklist
- [x] All tests still pass.
~~- [ ] Any new or changed features have are fully documented.~~

### TODO:
- [x] Write tests
- [x] Make more efficient
- [x] Consider caching groups somehow - could write a function that just takes the dictionary of template names and seed ids and uses and `lru_cache` on the result?


calum-chamberlain:
@flixha - this might be ready for you to have a play with and see if it helps with grouping your templates more sensibly, and reducing the nan-traces. It would be great to see if this improves your run-times. At the moment the first time you run the grouping for a new set of templates it is not very fast, and it scales poorly with number of templates (I tested with 5,000 templates and it took a few seconds to group). But it caches the two most recent results, so the next time you try to group, assuming you have the same templates, it will be near instantaneous.

Keen to hear your thoughts on the logic - it tries to group templates by the most similar seed-ids starting from the smallest template (fewest traces).

calum-chamberlain:
I had a play with this today to see if it made much difference to the current implementation and got slightly faster times overall (by about 5-10%) with most of the time saved doing the actual correlations because fewer redundant NaN correlations are run.  I'm keen to press on with this, but if anyone has any ideas for better ways to do this grouping then I would be happy to try them.

Otherwise I will work through the checklist tomorrow and aim to merge this tomorrow. With these speed-ups and dropping the pyfftw dependency I'm keen to get a new release out soon as well. (and that release will target up to py 3.11).

flixha:
Hi @calum-chamberlain,
thanks a lot for this work, this should really help gain some more performance and reducing memory load!
I have only been able to test this a tiny bit.
For checking the proper performance against the old implementation, I'll need to see when I have a system where performance isn't affected too much by other users..

From a first look at one case (re-checking of some hundred selected templates and detections out of a bigger dataset, so possibly very limited overlap between templates), it seems that in such cases the grouping may work over-zealously in splitting up the tribe into a lot of groups with mostly 1 template each; which then results in slower total times than previously (ca. 10 minutes vs 1 minute).
I haven't checked how exactly the grouping turns out in that case,in the new case each group takes about 2 seconds (11 minutes total) while in the old implementation it was more like 

From what I understood in the code it should already be able to match templates into one group as long as they have any overlap, so I'm not sure why it's splitting my test case into so many groups.
Maybe there could be a way to merge groups if there are too many that are too small?

calum-chamberlain:
Splitting into one template groups would be very slow! That isn't what I expected, nor what I observed with my template sets or tests - do you have a small set that shows that (ideally just the seed ids of the templates and stream so that I can make a small test case to catch and avoid that bug).
I wonder if this is happening when there is zero overlap in seed ids between templates?

calum-chamberlain:
@flixha - did you happen to have a case that had that strange grouping?

In other news, I haven't yet had time to put together a proper pull request with benchmarks, but you might be interested in trying the code in the `parallel-steps` branch... On my desktops and on our HPC system at VUW I get much better throughput on the GPU using FMF (still not a constant load, I think the only way we could achieve that would be handling memory allocation internally and finding peaks on the GPU, both of which are something for another branch).  There are lots of changes in there, and I need to do some tidying, but if you happen to be running some big template sets you might want to give that branch a test.

PR #552, opened 2023-10-27
Patches for several failing tests

calum-chamberlain:
### What does this PR do?

There are a few failing tests due to various upstream changes that need to be patched.


calum-chamberlain:
This PR updates the test subspace detectors. I think there have been upstream changes to SVD which have resulted in these changes. The detections made are the same as before, but the output for the SVD is different. Data in are the same so...

PR #570, opened 2024-05-13
Add py 3.12 to tests

calum-chamberlain:
### What does this PR do?

Add Python 3.12 to test suite and remove 3.8

Python 3.12 is being pushed out on conda-forge. We need to test this first here to make sure EQcorrscan runs correctly on 3.12.


calum-chamberlain:
Currently tests on Windows and MacOS are failing due to an issue sharing Client objects between processes. This looks like an upstream change, but I have opened an issue on obspy [here](https://github.com/obspy/obspy/issues/3528) to see if anyone else has run into this.

calum-chamberlain:
It looks like I will need to add a way to cope with the non-pickle-ability of obspy FDSN Clients. This could probably be handled by checking whether the client is an fdsn client and if it is passing the client kwargs (note, not just the url as other things might be needed) rather than the client, then instantiating a new client in each of the processes. This is probably the better way of doing it anyway.

PR #606, opened 2025-06-03
Patch 1

arif-geo:
### What does this PR do?
Fixes parallelization problem in clustering function (distance_matrix func)

I faced a weird problem- with increasing number of core, the runtime for clustering was increasing exponentially. cores=1 gave the lowest runtime. 
So I think the problem is in the deeply nested parallelization. So I talked to Copilot and edited the parallelization to be shallow and pass cores=1 for the deeper functions (only for clustering).
And runtime decreased significantly:
Old   code with 4 core: 109  sec
New code with 4 core: 26.3 sec

### PR Checklist
- [ ] `develop` base branch selected?
- [x] All tests still pass.

### DIFF
--- a/eqcorrscan/utils/clustering.py
+++ b/eqcorrscan/utils/clustering.py
@@ -12,6 +12,7 @@ import os
 import logging
 import ctypes
 from multiprocessing import cpu_count
+import concurrent.futures
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


calum-chamberlain:
Hi Arif,

Thanks for having a play with this. For this to be useful and accepted can you please provide a complete test example of what you were running with to test this. While the timings are useful to see, without knowing what you were actually running it is impossible to evaluate whether this speed up is useful for a general case.

